PR #22, opened 2024-10-10
Feature/14 implement map UI

misterM125:
**Summary**

This pull request introduces the map UI which is used to display a Google Map showing parks throughout the lausanne Canton via Markers

Navigation to other parts of the app is also partially implemented (related screens must be added later in MainActivity)

Screenshot from 2024-10-10 13-49-29
Screenshot from 2024-10-10 13-49-44

**To Be Implemented**
Here is a non-exhaustive list of features that are still in development and that have not been implemented on this branch yet:

- Markers should redirect to each park's respective event page
- navigation is not functional yet as other screens have not yet been implemented

**Testing**
To preview the map UI, you can change the `startDestination` of the `navhost` in the main composable to `Route.Map`:
`NavHost(
      navController = navController,
      startDestination =
          Route.MAP) `

(note that this will make navigation tests fail so change it back after you're done)


Simmanz (commented):
Hey.

- Overall the code is clean and clear to read.

**UX / UI :** 

- Google maps API works correctly and displays the map and pins on parks, user can move inside it (used temp API key).
- Clicking on the navigation menu's profile icon is indeed changing the screen, which is good.

**Tests :** 

- The `MapUiTest` covers the essential aspects of the MapScreen UI: visibility, interactions with the map and bottom navigation items. I like that it also tests for elements not to be shown before arriving to the Map screen (Auth screen), you could even check the same things for the other screens.

- We might need to add tests to verify the correct state management in the `ParkLocationViewModel `when the parks are implemented.

- Your code also need to have the needed 80% coverage on tests.

misterM125:
> Hey.
> 
> * Overall the code is clean and clear to read.
> 
> **UX / UI :**
> 
> * Google maps API works correctly and displays the map and pins on parks, user can move inside it (used temp API key).
> * Clicking on the navigation menu's profile icon is indeed changing the screen, which is good.
> 
> **Tests :**
> 
> * The `MapUiTest` covers the essential aspects of the MapScreen UI: visibility, interactions with the map and bottom navigation items. I like that it also tests for elements not to be shown before arriving to the Map screen (Auth screen), you could even check the same things for the other screens.
> * We might need to add tests to verify the correct state management in the `ParkLocationViewModel `when the parks are implemented.
> * Your code also need to have the needed 80% coverage on tests.

As it has now be clarified that the hard requirement is the line coverage testing now covers the requirements
Thank for your review!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
+#Mon Oct 07 10:26:18 CEST 2024
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
```
tercierp, line 4: Question: Are there any dependencies incompatible with gradle 8.4?

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Composable
+fun MapScreen(parkLocationViewModel: ParkLocationViewModel, navigationActions: NavigationActions) {
+
+  // hardcoded initial location values are used instead of the user's current location for now
+  val initialLatLng = LatLng(46.518659400000004, 6.566561505148001)
+
+  val parks = parkLocationViewModel.parks.collectAsState().value
+
+  parkLocationViewModel.findNearbyParks(initialLatLng.latitude, initialLatLng.longitude)
```
tercierp, line 35: Improvement: I think this line will leads to unnecessary recomputation of parks every time the composable is recomposed, you can use a `LaunchedEffect` to patch this problem :

```kotlin
LaunchedEffect(initialLatLng) {
    parkLocationViewModel.findNearbyParks(initialLatLng.latitude, initialLatLng.longitude)
}
```

Simmanz (approved):
As for the previous comment, I think that your PR satisfies the conventions and conditions we defined.

> Hey.
> 
>     * Overall the code is clean and clear to read.
> 
> 
> **UX / UI :**
> 
>     * Google maps API works correctly and displays the map and pins on parks, user can move inside it (used temp API key).
> 
>     * Clicking on the navigation menu's profile icon is indeed changing the screen, which is good.
> 
> 
> **Tests :**
> 
>     * The `MapUiTest` covers the essential aspects of the MapScreen UI: visibility, interactions with the map and bottom navigation items. I like that it also tests for elements not to be shown before arriving to the Map screen (Auth screen), you could even check the same things for the other screens.
> 
>     * We might need to add tests to verify the correct state management in the `ParkLocationViewModel `when the parks are implemented.
> 
>     * Your code also need to have the needed 80% coverage on tests.

.gitignore
```diff
@@ -1,12 +1,28 @@
-/.idea/
-.DS_Store
-.gradle
-/local.properties
-/build
-/captures
-.externalNativeBuild
-.cxx
-local.properties
```
tercierp, line 9: Question: Why do you removed some files in the gitignore ?
List of file removed:
- .gradle
- /local.properties
- /build
- /captures
- .externalNativeBuild
- .cxx
- local.properties

tercierp (changes requested):
As @Simmanz said very good code, I still have some questions about the gitignore changes and the gradle version change

gradle/wrapper/gradle-wrapper.properties
misterM125: Not sure, this seems to might have been updated automatically after using gradle commands, would there be a problem with using the most modern version anyway?

.gitignore
misterM125: I think this might just be due to a difference in placement for the files compared to the main branch, if you look at the gitignore on my branch, the files are still there

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
misterM125: Good call! This has been fixed, thanks.

gradle/wrapper/gradle-wrapper.properties
tercierp: We will keep this line, but in case of a Gradle build problem, this will be reverted.

tercierp (approved):

PR #23, opened 2024-10-10
Feature/12 implement sign up and log into the app

Simmanz:
## This branch consist of implementing the log-in UI screen and the sign-in logic.

- **SignInScreen** UI function in SignIn.kt
- **Utils folder** for helper functions.
- **AuthService interface** to abstract away authentication.
- **GoogleAuthService** is AuthService's implementation for google authentication.
- **LoginTest** which verifies that the SignInScreen UI works properly.
- **GoogleAuthServiceTest** which is a mock test for the GoogleAuthService.
- **Manual test** for the sign-in.
-  **Tweaks** to gradle / google-services / ressources.

### SignInScreen

1. **Basic UI elements** - Welcome text, Sign-in button with containers. New ressources for the google icon.
2. **Authentication logic**  - Component instantiation as well as lauching the sign-in (through the AuthService) on button click.
3. **Navigation** - Completing a sign-in with google navigates to the map screen (see launcher instantiation).

Screenshot_20241011_083226-2

### Utils folder

Contains the **AuthService** interface as well as the **GoogleAuthService** implementation

### AuthService interface
It has tree functions : 
- `launchSignIn (context, launcher)`- this function starts the sign-in process.
- `signOut ()` - this function signs the user out.
- `getCurrentUser () : FirebaseUser` - this function returns the currently logged-in user from firebase. This will need further abstraction in case we want new ways to authenticate.

### GoogleAuthService
- Implements the tree functions from the AuthService
- After some thinking, I chosed to also include the **rememberFirebaseAuthLauncher** composable because it felt unecessary to have it stay in the UI screen. It was to me against the idea of separating UI from sign-in logic.

Here are quotes from the bootcamp to further inform about this : 
> We recommend following [this tutorial](https://www.composables.com/tutorials/firebase-auth), which directly implements all the logic inside the composable. While this approach works, it's not considered best practice. Later, we will explore using the ViewModel and the MVVM pattern to separate the logic more effectively.

> You should isolate authentication behind an interface so that your tests for other components do not have to involve authentication directly, and can also test edge cases in users.

### LoginTest
This test focus on **UI testing** and used the following logic : 

- For **visible** elements such as Text, Image, Box - verify display.
- For **container** elements - verify existence.
- Verify **text values**.
- Verify **click action** on the button.
- Verify that **google responds** (packet) after a click is performed on the sign-in button.

Had to use `useUnmergedTree = true` for some checks because apparently compose puts some children elements (of a container like column/row) in a specific state which makes testing with testTags impossible. There are some other ways to circumvent this but I chose to use this way as it was the most straightforward to me.

### GoogleAuthServiceTest
Uses mockito and Junit to test the **GoogleAuthService**.
Every function is covered **except** for **rememberFirebaseAuthLauncher** because it is way harder to test and after hours of work I can still not get it working.

### Manual testing

Performed authentication with google through the **SignInScreen** UI. 
**Firebase** now has a registered user "_jwek0912@gmail.com_" (my testing account).

### Tweaks 

- Set up **SHA signature** for my side (on Firebase). This needs to be done by each person of the team.
- Added the **encrypted google-services** to the **github secrets**.
- Added the **default web client id** in `app/src/main/res/values/strings.xml`.
- Set up **gradle files** and **libs-versions**.
- Slight **reordering** of the gradle libraries / plugins.

### Current problems

- [x]  Did not find a way to test `GoogleAuthService's rememberFirebaseAuthLauncher`. (Most missing coverage is here)
- [x]  Tests no longer works after merging, probably an unintended dependency removal.
- [x]  CI was not checking for secrets, should be done but cannot check due to the problem with dependencies.
- [x] Problem with Intents / UI component not existing in UI tests.

### To test the feature

- You will need the `google-services.json` file from the Firebase project as it is unsing Firebase's Authentication.
- You will need to setup your **SHA_1** signature with Firebase.
- The navigation has start screen set up to `Screen.AUTH`, meaning simply running the `StreetWorkAppMain()` will run `SignInScreen()`.


app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
```diff
@@ -0,0 +1,131 @@
...
+              onClick = {
+                Log.d("SignInScreen", "Start sign-in")
+                authService.launchSignIn(context, launcher)
+              },
+              modifier =
+                  Modifier.width(250.dp)
+                      .height(40.dp)
+                      .border(
+                          width = 1.dp,
+                          color = Color(0xFFDADCE0),
```
Alex720p, line 96: Suggestion: don't use a hardcoded color and define it in a color palette somewhere with a proper name

app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
```diff
@@ -0,0 +1,131 @@
...
+                          shape = RoundedCornerShape(20.dp))
+                      .background(color = Color(0xFFFFFFFF), shape = RoundedCornerShape(20.dp))
+                      .testTag("loginButton"),
+              colors =
+                  ButtonDefaults.buttonColors(
+                      containerColor = Color.White, // Custom background color
+                      contentColor = Color(0xFF3C4043) // Custom text color
```
Alex720p, line 103: same here for the color palette

app/src/test/java/com/android/streetworkapp/auth/GoogleAuthServiceTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    // Act
+    val currentUser = googleAuthService.getCurrentUser()
+
+    // Assert
+    assertEquals(mockUser, currentUser) // Check that the returned user is the same as the mock user
+  }
+
+  @Test
+  fun testGetCurrentUserReturnsNullWhenNuserIsSignedIn() {
```
Alex720p, line 50: Improvement: typo in test name

app/src/test/java/com/android/streetworkapp/auth/GoogleAuthServiceTest.kt
```diff
@@ -0,0 +1,60 @@
...
+    // Arrange
+    `when`(mockAuth.currentUser).thenReturn(null)
```
Alex720p, line 52: error: this test doesn't do what it's supposed to do

app/src/androidTest/java/com/android/streetworkapp/ui/authentication/LoginTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    // For columns / rows / spacers : check if exist :
+    composeTestRule.onNodeWithTag("loginScreenColumnContainer").assertExists()
+    composeTestRule.onNodeWithTag("loginScreenSpacer").assertExists()
+    composeTestRule.onNodeWithTag("loginScreenSpacer").assertExists()
+    composeTestRule.onNodeWithTag("loginButtonRowContainer", useUnmergedTree = true).assertExists()
+
+    // UX elements :
+
+    // UX - Text values :
+    composeTestRule.onNodeWithTag("loginTitle").assertTextEquals("Welcome to the Street Work'App")
```
Alex720p, line 54: Improvement: I guess this is to gain coverage but checking for text values isn't the best idea. Would need to change the test if we decide to change the text etc... Would be better to only check if it exists since it's more flexible

app/src/androidTest/java/com/android/streetworkapp/ui/authentication/LoginTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+  @get:Rule val composeTestRule = createAndroidComposeRule<MainActivity>()
+
+  // The IntentsTestRule simply calls Intents.init() before the @Test block
+  // and Intents.release() after the @Test block is completed. IntentsTestRule
+  // is deprecated, but it was MUCH faster than using IntentsRule in our tests
+  @get:Rule val intentsTestRule = IntentsTestRule(MainActivity::class.java)
+
+  @Test
+  fun UIComponentsDisplayed() {
```
Alex720p, line 30: Improvement: test name should start with lowercase

Alex720p (changes requested):
very good overall but there are some things I would change:
- don't hardcode values, either define them or use some already defined
- there is an error in a test, it should be redone
- there is probably some redundancy with .assertExists().assertIsDisplayed()

app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
Andy130604, line 1: Small note for everyone, when writing UI code in jetpack compose, try limiting your composable functions in length. Create smaller, maybe more reusable composables and don't just put everything in a 'screen'.

app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
Andy130604: This is a great remark!

app/src/androidTest/java/com/android/streetworkapp/ui/authentication/LoginTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    // Test uses useUnmerged = true for all children of containers,
+    // otherwise will not be accessible for the test using testTags.
+
+    // For Box, Text, Image, Buttons, List : check if displayed :
+    composeTestRule.onNodeWithTag("loginScreenBoxContainer").assertExists().assertIsDisplayed()
```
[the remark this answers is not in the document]
Simmanz: It is not, because in some cases where a component is a children of another (for instance inside a container like col / row), then it might be put in the unmerged tree and the assertIsDisplayed will return false, while the assert exists will return true. I put this check in case it happens to enable quick fixing.

app/src/androidTest/java/com/android/streetworkapp/ui/authentication/LoginTest.kt
Simmanz: I assumed that because the Figma is part of the grade, we need to have an UI that matches what is described there. And text values where part of it.

app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
Simmanz: This was corrected in recent fix of using the Color.kt file of the project

app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
Simmanz: I abstracted away the Google Authenthication button in a separate file to enhance readability / reusability. 

I also tried to separate the launcher instantiation in another function but did not manage it yet due to the fact that even as a mutableState, the user variable cannot be modified after being passed as an argument to the abstracted function (it is considered to be a val). Is it too much to do this also ?

Alex720p (approved):

PR #27, opened 2024-10-12
Feat/GitHub release integration

tercierp:
This github action will create an apk version of our application automatically each time a branch is merged into main. You can also run this action by going to the actions section of github.

This could be tested when the branch is merge with the main


.github/workflows/apk-builder.yml
```diff
@@ -0,0 +1,93 @@
+name: APK build
+on:
```
Andy130604, line 2: Don't add the on push to main functionality. Just let it be a workflow dispatch, it's fine. Otherwise you will use the github runners for nothing most of the time + using artifacts space for nothing too.

tercierp: You are right fixed ✅

.github/workflows/apk-builder.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      # Load the google-services.json file from the secrets
+      - name: Load Google Service file
+        env:
+          DATA: ${{ secrets.GOOGLE_SERVICES_JSON }}
```
misterM125, line 26: typo: the current secret for the google services json file is called GOOGLE_SERVICES and not GOOGLE_SERVICES_JSON

.github/workflows/apk-builder.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          name: streetworkapp
+          path: app/build/outputs/apk/release/${{ steps.tag.outputs.release_tag }}.apk
+          if-no-files-found: error
+          retention-days: 90
+
+      - name: Load Google Maps API Key
+        env:
+          MAPS_API_KEY: ${{ secrets.MAPS_API_KEY }}
+        run: echo "MAPS_API_KEY=${MAPS_API_KEY}" >> local.defaults.properties
```
misterM125, line 74: fix: the google maps API key is currently stored in the LOCAL_PROPERTIES secret, furthermore it is stored in base 64 so you might have to decode it before using it

.github/workflows/apk-builder.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Load Firebase API Key
+        env:
+          FIREBASE_API_KEY: ${{ secrets.FIREBASE_API_KEY }}
+        run: echo "FIREBASE_API_KEY=${FIREBASE_API_KEY}" >> local.defaults.properties
```
misterM125, line 79: Fix: for confidentiality purposes, the firebase API key will more than likely also be encoded in base 64 before being posted on github secrets, so the APK might need to decode it before using it

misterM125 (changes requested):
Overall the APK release looks very good and functional, there are however a few minor problems that might not make it work such as not properly decoding the secrets before using them or not using the right names, which might be something that we might change either directly in code or on github, I will let you be the judge of that 

Thank you for your PR!

misterM125:
**additional note**

I think as a convention we might encode all secrets on github in base 64, so it might be mindful of always decoding them before using them

.github/workflows/apk-builder.yml
tercierp: Question: are you sure that encoding in base64 provides confidentiality? From my perspective, I do not think so. Based on [this source](https://developer.mozilla.org/en-US/docs/Glossary/Base64), we can see that the encoding is reversible.

But I did remove the Firebase key

.github/workflows/apk-builder.yml
tercierp: Yep done ✅

.github/workflows/apk-builder.yml
misterM125: Apologies, I thought that was the reason we encoded it in the bootcamp but after rereading the assignment i think it is only done for us to easily store files as secrets on github (since github can only store text secrets)
I still think we should store all our secrets encoded in base 64 to avoid any confusion between those who are and those who are not

misterM125 (approved):
looks good to me,
Thank you for your changes!

PR #37, opened 2024-10-16
Feat/UI to create new events

alvaro1080:
**An UI to create new events**

The objective of this pull request is to add a new screen that will be used to create and add a new event to a given park. I tried as much as possible to make sure that the interface is easily usable on a smartphone. This is the reason why a slider is used for the selection of the maximum number of participants of an event. The selection of the date selection is improved to also allow the user to pick a specific hour of the day, since it is important to have a clear hour for the events.

The UI is not totally functional yet, as it will need a viewModel for the parks and one for the events in order to be able to use Firebase. Those viewModels should be available soon.

**Features**
- two text field to specify the name and the description of the event
- A "time selection" composable, that can be used to first select the date, and then the hour of the event
- A slider to select the maximum number of participants

**Objective of the PR**
In addition of any improvement you might suggest, I would really appreciate a feedback about the user experience. In particular, do you think that this UI is easily usable on a smartphone? (If anyone could try on a real device, I think it could be useful)

**Screenshots**
image

image

image

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
+++ b/app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
@@ -1,21 +1,51 @@
 package com.android.streetworkapp.ui.park
 
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.AddCircle
+import androidx.compose.material.icons.filled.DateRange
+import androidx.compose.material3.Button
+import androidx.compose.material3.DatePicker
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.OutlinedTextField
 import androidx.compose.material3.Slider
 import androidx.compose.material3.SliderDefaults
 import androidx.compose.material3.Text
+import androidx.compose.material3.TimePicker
+import androidx.compose.material3.rememberDatePickerState
+import androidx.compose.material3.rememberTimePickerState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableFloatStateOf
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.shadow
 import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.window.Popup
 import com.android.streetworkapp.model.event.Event
+import com.google.firebase.Timestamp
+import java.text.SimpleDateFormat
+import java.util.Calendar
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+import java.util.concurrent.TimeUnit
 
 /** Display a view that is used to add a new Event to a given park. */
 @Composable fun AddEventScreen() {}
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
```diff
@@ -0,0 +1,305 @@
...
+  Column(
+      verticalArrangement = Arrangement.Center,
+      horizontalAlignment = Alignment.CenterHorizontally,
+      modifier = Modifier.fillMaxWidth()) {
+        Slider(
+            modifier = Modifier.fillMaxWidth(0.8f).testTag("sliderMaxParticipants"),
+            value = sliderPosition,
+            onValueChange = {
+              sliderPosition = it
+              event.maxParticipants = sliderPosition.toInt()
```
Alex720p, line 182: Suggestion: (this **might** be a non issue but it will depend on our later implementation.) In this composable and other composables you immediately change the value of the event on each input. However data classes are passed by reference. If someone goes in, makes changes, doesn't want to save and goes back, it could lead to the object he has locally not matching the one in the database if we don't resync. (Again this will depend on our later implementations but I think it can be dangerous)

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
```diff
@@ -0,0 +1,305 @@
...
+}
+
+/**
+ * Used to change the maximum number of participants of the event
+ *
+ * @param event the event that will be updated
+ */
+@Composable
+fun ParticipantNumberSelection(event: Event) {
+  val minParticipants = 2 // We need at least 2 users for every event
```
Alex720p, line 169: Improvement: These 2 shouldn't be here, they should be put in another file so that we can easily access then inside other components. (I think these values are already hardcoded in some other tests that we should change but that's beside the scope of this PR)
 They should also be renamed to MIN_PARTICIPANTS AND MAX_PARTICIPANTS to make it clear that these values are constants.

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
```diff
@@ -0,0 +1,305 @@
...
+          Row {
+            IconButton(
+                modifier = Modifier.testTag("dateIcon"),
+                onClick = { showDatePicker = !showDatePicker }) {
+                  Icon(imageVector = Icons.Default.DateRange, contentDescription = "Select date")
+                }
+            IconButton(
+                modifier = Modifier.testTag("timeIcon"),
+                onClick = { showTimePicker = !showTimePicker }) {
+                  Icon(imageVector = Icons.Default.AddCircle, contentDescription = "Select time")
```
Alex720p, line 240: Suggestion: I think this icon is a bit misleading, I couldn't tell it was for the time at first glance. Should probably be changed to a clock or something similar to remove ambiguities

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
```diff
@@ -0,0 +1,305 @@
...
+        Box(
+            modifier =
+                Modifier.fillMaxWidth()
+                    .shadow(elevation = 4.dp)
+                    .background(MaterialTheme.colorScheme.surface)
+                    .padding(16.dp)) {
+              DatePicker(state = datePickerState, showModeToggle = false)
+            }
+
+        Button(
```
Alex720p, line 257: Improvement: I can validate dates that are earlier in time (e.g October 2020), we'll probably do the checks in the viewmodel and whatnot anyways but it would be a good addition to the UI that a user can't validate an invalid date

Alex720p (changes requested):
Really clean code, tests and UI!
However there's a few things that need change:
 - modify the naming and place of the constants
 - review how the events are modified or be clear with the rest of group how we pass events to the composable (i.e by copy if we keep the code like this)
 - Change the logo for the time input to remove ambiguities
 - Make it so the UI informs the user when an invalid date is used as input

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
alvaro1080: I understand your concern. I think that, in the context of this screen, this is not a problem since the objective is only  create locally a new Event, that will later be sent to the database. This is like a "builder" for events, that once this event is ready locally, is sent to the database. This means that there should not be a situation where an event is fetch from the database (in this screen at least), since it is only used to add new events (this would be expected probably on an "edit event" screen perhaps). Therefor, I think that having an event as an argument for every composable, that is always an event created locally to be edited by the user through the UI and then sent to the database once it is ready, is the best way to have small and understandable composable. In this context, the "pass by reference" is intended.

This is why I think that we should keep this implementation. Do you agree?

Alex720p: Yeah all good, I mixed things up and thought it would also do edits for some reason. my bad.

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
alvaro1080: You are right, I defined with an Object "EventConstants" in the Event.kt file. Thank you!

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
alvaro1080: You are right, this should be improved. Unfortunately, this task is harder than it seems. Do you mind if we do this update in another future sprint?

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
alvaro1080: I agree with you. There was no clock icon in the basic set of icon so I had to import "libs.androidx.material.icons.extended". Thank you!

app/src/main/java/com/android/streetworkapp/ui/park/AddEvent.kt
Alex720p: Yep let's do that

Alex720p (approved):

PR #39, opened 2024-10-17
Feat/link composables with navigation

Alex720p:
- **ParkOverview TopBar:**
  - Created the TopBar in the ParkOverview Composable to easily navigate back to the map
  - Implemented tests for this new element in the already existing `ParkOverviewTest.kt` file
  - Added `ParkOverviewScreen` prefix to the old tests in `ParkOverviewTest.kt` to remove ambiguities
 
- **Navigation:**
  - Defined a new screen `Park Overview Screen` for our new ParkOverview Composable

- **MainActivity:**
  - Added the StreetWorkAppMain composable, this allows us to create our repos, viewmodels etc... before calling StreetWorkApp. This allows for easy injecting of mocks in our tests
  - Updated the hierarchy to include the ParkOverview Composable inside our `Map` route 

- **End2End Tests:**
  - Added a new parameter `callbackOnMapLoaded` that gets called once the GoogleMap Composable has finished loading, its default value is {}
  - Added a file `End2EndParks.kt` containing a new `E2ECanClickParkAncAccessSpecificParkOverview` test. **Note**: This test has a `@Ignore` tag to it as sometimes Google Map can take some time to load and make the test fail (very rare but can happen). This test is designed to run locally as one wouldn't want the CI to fail because of an external API.

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/MainActivity.kt
+++ b/app/src/main/java/com/android/streetworkapp/MainActivity.kt
@@ -9,16 +9,25 @@ import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
 import androidx.navigation.navigation
+import com.android.streetworkapp.model.event.Event
+import com.android.streetworkapp.model.event.EventList
+import com.android.streetworkapp.model.park.Park
+import com.android.streetworkapp.model.parklocation.OverpassParkLocationRepository
+import com.android.streetworkapp.model.parklocation.ParkLocationViewModel
+import com.android.streetworkapp.ui.map.MapScreen
 import com.android.streetworkapp.ui.navigation.NavigationActions
 import com.android.streetworkapp.ui.navigation.Route
 import com.android.streetworkapp.ui.navigation.Screen
+import com.android.streetworkapp.ui.park.ParkOverview
 import com.android.streetworkapp.ui.profile.ProfileScreen
+import com.google.firebase.Timestamp
+import okhttp3.OkHttpClient
 
 class MainActivity : ComponentActivity() {
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     Log.d("MainActivity", "Setup content")
-    setContent(parent = null) {}
+    setContent(parent = null) { StreetWorkAppMain() }
   }
 }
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/streetworkapp/MainActivity.kt
```diff
@@ -9,26 +9,75 @@ import androidx.navigation.compose.NavHost
...
+          location = "EPFL",
+          image = null,
+          rating = 4.5f,
+          nbrRating = 102,
+          occupancy = 0.8f,
+          events = eventList)
+
   NavHost(
       navController = navController,
       startDestination = Route.AUTH) { // TODO: handle start destination based on signIn logic
```
alvaro1080, line 83: Proposition: In order to have a runnable demo for the next coaching meeting, I think it could be useful to have a temporary starting screen, to make sure that the main is directly runnable. 
Here is a simple exemple:

```
composable(Screen.AUTH) {
            Column(
                modifier = Modifier.fillMaxSize(),
                verticalArrangement = Arrangement.Center,
                horizontalAlignment = Alignment.CenterHorizontally) {
                  Button(onClick = { navigationActions.navigateTo(Screen.MAP) }) { Text("Start") }
                }
          }
```

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -52,14 +57,18 @@ fun MapScreen(parkLocationViewModel: ParkLocationViewModel, navigationActions: N
 
         // Display the Google Map
         GoogleMap(
+            onMapLoaded = callbackOnMapLoaded,
```
alvaro1080, line 60: Question: what is "callbackOnMapLoaded" used for?

alvaro1080 (changes requested):
Thank you for your PR! The navigation does work on my device, and everything that has been added until now is correctly linked to each other. I think we are ready to provide a demo with our progress until today with your Pull Request, and that is great! I have a few points that I would like to discuss:

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
Alex720p: for the e2e test, it decreases the counter of an idlingRessource that deblocks the test and makes it resume. Otherwise you can't click on Markers as they aren't loaded

app/src/main/java/com/android/streetworkapp/MainActivity.kt
Alex720p: Good idea, I implemented this but changed `navigationActions.navigateTo(Screen.MAP)` to `navigationActions.navigateTo(Route.MAP)` to clear the backstack

PR #42, opened 2024-10-17
Feat/implement add friend.UI

kzepfl:
## Changes

- Implemented UI for AddFriendScreen (`AddFriend.kt`)

_The NFC don't work yet, so we try to first add friend using ID
The button `send request` should add the ID to the friendlist_

- Added AddFriendScreen to Navigation (`MainActivity.kt` and `NavigationActions.kt`)

## Testing

- Implemented UI tests for the AddFriendScreen composable. (`AddFriendScreenTest.kt`)

_The test verify that all element are displayed and the button when clicked correctly add friends_


app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
```diff
@@ -0,0 +1,114 @@
...
+                    } else {
+                      // add the friend to the user firendlist
+                      userViewModel.addFriend(uid, id)
+                      Toast.makeText(context, "Friend request sent.", Toast.LENGTH_SHORT).show()
+                      /*try {
+                        userViewModel.addFriend(uid, id)
+                        Toast.makeText(context, "Friend request sent.", Toast.LENGTH_SHORT).show()
+                      } catch (e: ApiException) {
+                        Toast.makeText(context, "Invalid ID", Toast.LENGTH_SHORT).show()
+                      }*/
```
Simmanz, line 106: This should be removed, it is against conventions to let junk code commented out.

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
```diff
@@ -0,0 +1,114 @@
...
+              Image(
+                  painter = painterResource(id = R.drawable.place_holder),
+                  contentDescription = "profile picture",
+                  modifier = Modifier.size(200.dp))
+
+              // button to activate NFC (don't work)
+              Button(onClick = {}, modifier = Modifier.size(220.dp, 50.dp).testTag("NFCButton")) {
+                Text(text = "Activate NFC", fontSize = 17.sp)
+              }
+
```
Simmanz, line 80: You could separate this code from the UI screen by putting it in a @Composable function at the bottom of the file. This is important and was given as advice from the coaches in my branch :

> Small note for everyone, when writing UI code in jetpack compose, try limiting your composable functions in length. Create smaller, maybe more reusable composables and don't just put everything in a 'screen'.

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
Simmanz, line 1: Code is clear and well-commented, one instance of commented-out code should be removed though. Could benefit from getting some components out to improve further. Otherwise good !

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz, line 1: Clear changes, had to update signature due to new merge to main and added navigation to make it possible to access the ui.

app/src/androidTest/java/com/android/streetworkapp/ui/profile/AddFriendScreenTest.kt
Simmanz, line 1: Clear checks on the display of UI elements.
Also really good that you test that adding a friend with empty ID is blocked.
You also verify that a correct call to the MVVM is made when adding friends through the ID.

Simmanz (changes requested):
### Overall clear code and tests.

-Navigation works.
-Use of the user MVVM model.
-Testing of UI display and edge cases.

### Could benefit from some improvements :

- Some commented-out code remains
- Your UI could be improved by defining some elements outside of the UI screen code (end of file) in the way coaches suggested (see comment inside code).

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
kzepfl: ????

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
kzepfl: i think making a compodable for 3 line of code for a button is too extra

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: I meant that this was expected things to do.

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
Simmanz: Yes it seems extra, but when we will continue to implement this button to handle NFC, we will be happy that it is separated from the ui screen to prevent having a huge function (the main argument of the coach). 

This is the same case as with the `GoogleAuthButton`, and it would be important to do the same.

Simmanz (approved):

PR #43, opened 2024-10-17
Hotfix/user mvvm

tercierp:
This PR addresses several issues in the MVVM architecture:
- Created the `setCurrentUser` function.
- Renamed the `getUserById` function.
- Added automatic creation of a Firebase `user` in `SignIn.kt`.
- Modified the signatures of some functions in the `ViewModel`.


app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
```diff
@@ -46,7 +49,13 @@ fun SignInScreen(navigationActions: NavigationActions, userViewModel: UserViewMo
...
           onAuthComplete = { result ->
             user = result.user
-            // todo call setCurrentUser in the viewmodel : userViewModel
+            userViewModel.viewModelScope.launch {
+              val fetchedUser = userViewModel.getUserByUid(user!!.uid)
+              if (fetchedUser == null) {
+                val newuser = User(user!!.uid, user!!.displayName!!, user!!.email!!, 0, emptyList())
+                userViewModel.addUser(newuser)
+              }
+            }
```
Simmanz, line 58: It might be good to encapsulate the logic inside of a function to minimize the amount of code in the UI. Because none of the values defined there will be used except the viewModel.

For instance either in the userViewModel or the bottom of the SignIn.kt file. 
A simple signature taking the user (and maybe teh viewModel depending on where you do it) could do the job.

app/src/test/java/com/android/streetworkapp/model/user/UserViewModelTest.kt
Simmanz, line 1: Changes are also clear here, you renamed the function in order to prevent misunderstandings.

Simmanz (changes requested):
### Overall very clear code and intent. 

The sole change that could be done is to abstract away the added lines in the `onAuthComplete` of the launcher such that it is cleaner in the UI. 

(Also we could re-use it _in case the firebase auth behavior does not go through the launcher again when an already signed-in user starting the process_. Which would (in hypothesis) introduce a discontinuity between the current user and the repository / MVVM).

app/src/main/java/com/android/streetworkapp/ui/authentication/SignIn.kt
tercierp: Done I let you check the changes and mark as resolved

app/src/main/java/com/android/streetworkapp/model/user/UserViewModel.kt
Simmanz, line 1: ### Unsure about the usage of `.postvalue()` - maybe `setvalue()` instead ?

Looked into a [post](https://stackoverflow.com/questions/51299641/difference-of-setvalue-postvalue-in-mutablelivedata) about it which told that `.setvalue()` should be used for main CPU treads and that `postvalue()` will have the last call be the only one to set a value. Also it is overriden by `.setvalue()` and is slower to run.

Taking into account the fact that android compose does not uses threading by default and that we might call get multiple users in a small time interval in the future, we could use the change.

What do you think about this ?

### Function enhancement (optional - if possible)

Could it be possible for the `getUser` functions to also return the user ? I think it could be useful because it would prevent needing two calls for essentially one getter.

Or would this bring too many issues with the current implementation and be against the current way of accessing using the LiveData ?

Simmanz (changes requested):
### Overall agreement - questions remains

1. Unsure about the usage of the postvalue() method (see comment for details).
2. Could the `getUser` methods return the user for simplicity, or is it better with two calls for one getter ?

That appart your tests make sense and good code quality. Also the code for user MVVM handling has indeed been abstracted out.

app/src/main/java/com/android/streetworkapp/model/user/UserViewModel.kt
tercierp: > ### Unsure about the usage of `.postvalue()` - maybe `setvalue()` instead ?
> Looked into a [post](https://stackoverflow.com/questions/51299641/difference-of-setvalue-postvalue-in-mutablelivedata) about it which told that `.setvalue()` should be used for main CPU treads and that `postvalue()` will have the last call be the only one to set a value. Also it is overriden by `.setvalue()` and is slower to run.
> 
> Taking into account the fact that android compose does not uses threading by default and that we might call get multiple users in a small time interval in the future, we could use the change.
> 
> What do you think about this ?

You are right ! good catch



> Or would this bring too many issues with the current implementation and be against the current way of accessing using the LiveData ?

Indeed there is some consistency issues, it’s best to stick with `LiveData` alone. This keeps the code consistent and enables the UI to automatically refresh without additional handling.

Simmanz: > Indeed there is some consistency issues, it’s best to stick with `LiveData` alone. This keeps the code consistent and enables the UI to automatically refresh without additional handling.

Alright, thanks for the information !

PR #47, opened 2024-10-24
feat/mvvm-get-events

misterM125:
**Overview**
This PR aims to implement the functionalities to fetch events from the firestore database to display them from a park overview screen

**features**
- getEvents has been created in the event viewModel to fetch events from the firestore database
- an EventOverviewUistate has been created to switch between a park overview screen with no events and one with
- EventviewModel has been added as a parameter to the park overview screen
- the park overview screen has been modified to take into account the viewModel and will now display events fetched from the firestore database

**Additionnal notes**

- As the park mvvm has not been fully implemented as of this PR the viewModel will display all events stored on firebase instead of events related to a specific park, this will need to be changed later one
- the "about" button does not yet redirect to an selected event's screen, this should be handled in another branch

**testing**
To test this new feature, launch the mainactivity with map as start destination and click on a marker on the map which should redirect you the park overview screen.
If some events are stored on firebase, you should be able to see some events displayed on the screen, else you can create an event  thanks to the "create an event" button which should then be added to the list of events on the park overview screen

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/model/event/EventViewModel.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/event/EventViewModel.kt
@@ -1,10 +1,18 @@
 package com.android.streetworkapp.model.event
 
+import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
 
 open class EventViewModel(private val repository: EventRepository) : ViewModel() {
+
+  private val _uiState: MutableStateFlow<EventOverviewUiState> =
+      MutableStateFlow(EventOverviewUiState.Empty)
+
+  val uiState: StateFlow<EventOverviewUiState> = _uiState
 
   /**
    * Get a new event ID.
@@ -15,10 +23,29 @@ open class EventViewModel(private val repository: EventRepository) : ViewModel()
     return repository.getNewEid()
   }
 
+  /** Fetch all events from the database. */
+  fun getEvents() {
+    repository.getEvents(
+        onSuccess = {
+          if (it.isEmpty()) {
+            _uiState.value = EventOverviewUiState.Empty
+          } else {
+            _uiState.value = EventOverviewUiState.NotEmpty(it)
+          }
+        },
+        onFailure = { Log.e("FirestoreError", "Error getting events: ${it.message}") })
+  }
+
   /**
    * Add a new event to the database.
    *
    * @param event The event to add.
    */
   fun addEvent(event: Event) = viewModelScope.launch { repository.addEvent(event) }
+}
+
+sealed class EventOverviewUiState {
+  data class NotEmpty(val eventList: List<Event>) : EventOverviewUiState()
+
+  data object Empty : EventOverviewUiState()
 }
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
```diff
@@ -271,20 +271,20 @@ fun DashBoardBar() {
...
 }
 
 /** Represents the different states of the event dashboard */
-sealed class OverviewUiState {
-  data object Details : OverviewUiState()
+sealed class DashboardState {
+  data object Details : DashboardState()
 
-  data object Participants : OverviewUiState()
+  data object Participants : DashboardState()
```
alvaro1080, line 289: Question: Why did this sealed class changed?

app/src/main/java/com/android/streetworkapp/model/event/EventViewModel.kt
```diff
@@ -1,11 +1,19 @@
...
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
 
 open class EventViewModel(private val repository: EventRepository) : ViewModel() {
 
+  private val _uiState: MutableStateFlow<EventOverviewUiState> =
+      MutableStateFlow(EventOverviewUiState.Empty)
+
+  val uiState: StateFlow<EventOverviewUiState> = _uiState
+
```
alvaro1080, line 16: Question: I am unsure about the use of EventOverviewUiState, is there an advantage to use it instead of a List<Event>? The name of the attribute (uiState) could also become confusing, do you think we could perhaps change it to a more specific name?

app/src/main/java/com/android/streetworkapp/model/event/EventViewModel.kt
```diff
@@ -15,10 +23,29 @@ open class EventViewModel(private val repository: EventRepository) : ViewModel()
...
    * @param event The event to add.
    */
   fun addEvent(event: Event) = viewModelScope.launch { repository.addEvent(event) }
 }
+
+sealed class EventOverviewUiState {
+  data class NotEmpty(val eventList: List<Event>) : EventOverviewUiState()
+
+  data object Empty : EventOverviewUiState()
+}
```
alvaro1080, line 51: (second part of the previous comment): As mentioned in the last comment, I am not sure that using this class has an advantage over using just the List <Event>, is there a reason to keep it?

alvaro1080 (approved):
Thank you for this Pull Request! After trying the test you suggested, the events are correctly displayed and are correctly fetched from the database. The code is clear, I only have a few concerns I would like to discuss:
(Note: since my comments only concern minor aspects and some questions, I have already approved your Pull request)

app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
misterM125: I changed the name of the class to not confuse the overview ui states with the dasboard ui state, i think the naming makes it more clear

app/src/main/java/com/android/streetworkapp/model/event/EventViewModel.kt
misterM125: Could you maybe clarify what you find confusing about this? 
I use a sealed class first to define the states of our overview screen (empty vs not empty) which i think makes the code more readable and also to make it that i don't necessarily have to fetch the EventList from the viewModel which makes the code more clean as it reduces the amount of public variables or potential calls to the `getEvents` method

alvaro1080: This was clarified in person, thank you for your explanation! (I therefor will mark this as resolved)

PR #53, opened 2024-10-29
Feat/implement single top level scaffold

Alex720p:
- **New Navigation classes and composable:**
  - Created  TopBarWrapper composable
  - Created ScreenParams and TopBarManager class that easily allows us to manage the screens bottom and top bar properties
  - Created a companion onject in ScreenParams that stores all the screen properties, this allows for topbar modification from any file/composable

- **MainActivity:**
  - Defined a single top level scaffold that shows the top/bottom bar depending on the current screen properties

- **Old Screens/Composables:**
  - Changed all the screens (except for EventOverview as it is too time consuming and should require a single issue for this task alone) to no longer have a scaffold
  - Adapted the existing tests of those composable to be compatible with the new structure

- **New Tests:**
  - Refactored the tests in BottomNavigationTest to test our new single top level scaffold structure
  - Implemented TopAppBarTest to test our new single top level scaffold new structure

### DIFF
--- a/app/src/androidTest/java/com/android/streetworkapp/ui/navigation/TopAppBarTest.kt
+++ b/app/src/androidTest/java/com/android/streetworkapp/ui/navigation/TopAppBarTest.kt
@@ -8,6 +8,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.ui.test.assertCountEquals
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsNotDisplayed
+import androidx.compose.ui.test.assertTextEquals
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onAllNodesWithTag
 import androidx.compose.ui.test.onNodeWithTag
@@ -70,7 +71,15 @@ class TopAppBarTest {
         parkLocationViewModel = ParkLocationViewModel(parkLocationRepository)
     }
 
-    
+    @Test
+    fun changingTitleInManagerMakesItChangeOnScreen() {
+        val topAppBarManager = TopAppBarManager("old title")
+
+        topAppBarManager.setTopAppBarTitle("new title")
+        composeTestRule.setContent { TopAppBarWrapper(NavigationActions(mockk()), topAppBarManager) }
+        composeTestRule.onNodeWithTag("topAppBarTitle").assertIsDisplayed().assertTextEquals("new title")
+    }
+
     @Test
     fun isDisplayedCorrectlyOnScreens() {
         val currentScreenParam = mutableStateOf(LIST_OF_SCREENS.first()) //can't call setContent twice per test so we use this instead
--- a/app/src/androidTest/java/com/android/streetworkapp/ui/navigation/TopAppBarTest.kt
+++ b/app/src/androidTest/java/com/android/streetworkapp/ui/navigation/TopAppBarTest.kt
@@ -1,0 +1,105 @@
+package com.android.streetworkapp.ui.navigation
+
+import android.annotation.SuppressLint
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.ui.test.assertCountEquals
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertIsNotDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onAllNodesWithTag
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onRoot
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.printToLog
+import com.android.streetworkapp.StreetWorkApp
+import com.android.streetworkapp.model.event.EventRepositoryFirestore
+import com.android.streetworkapp.model.event.EventViewModel
+import com.android.streetworkapp.model.park.ParkRepositoryFirestore
+import com.android.streetworkapp.model.park.ParkViewModel
+import com.android.streetworkapp.model.parklocation.OverpassParkLocationRepository
+import com.android.streetworkapp.model.parklocation.ParkLocation
+import com.android.streetworkapp.model.parklocation.ParkLocationViewModel
+import com.android.streetworkapp.model.user.UserRepositoryFirestore
+import com.android.streetworkapp.model.user.UserViewModel
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+
+class TopAppBarTest {
+
+    private lateinit var parkLocationRepository: OverpassParkLocationRepository
+    private lateinit var parkLocationViewModel: ParkLocationViewModel
+
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+
+    @Before
+    fun setUp() {
+        val mockParkList =
+            listOf(
+                ParkLocation(
+                    lat = 46.518659400000004,
+                    lon = 6.566561505148001,
+                    id = "1"),
+                ParkLocation(lat = 34.052235, lon = -118.243683, id = "2"),
+                ParkLocation(lat = 51.507351, lon = -0.127758, id = "3"),
+                ParkLocation(lat = 35.676192, lon = 139.650311, id = "4"),
+                ParkLocation(lat = -33.868820, lon = 151.209290, id = "5")
+            )
+
+        parkLocationRepository =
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
```diff
@@ -39,7 +35,8 @@ import com.android.streetworkapp.ui.navigation.NavigationActions
 @Composable
 fun AddFriendScreen(
     navigationActions: NavigationActions,
-    userViewModel: UserViewModel = viewModel(factory = UserViewModel.Factory)
+    userViewModel: UserViewModel = viewModel(factory = UserViewModel.Factory),
+    innerPaddingValues: PaddingValues = PaddingValues(0.dp)
```
tercierp, line 39: Question: same question as the previous padding problem what's the use case ?

app/src/androidTest/java/com/android/streetworkapp/ui/navigation/TopAppBarTest.kt
```diff
@@ -0,0 +1,100 @@
...
+
+      composeTestRule.waitForIdle()
+      if (screenParam.isTopBarVisible) {
+        composeTestRule.onNodeWithTag("topAppBar").assertIsDisplayed()
+        screenParam.topAppBarManager?.let { topAppBarManager ->
+          if (topAppBarManager.hasNavigationIcon())
+              composeTestRule.onNodeWithTag("goBackButton").assertIsDisplayed()
+        }
+      } else composeTestRule.onNodeWithTag("topAppBar").assertIsNotDisplayed()
+    }
```
tercierp, line 98: Note: no need to change this code but if these lines were in the code of our app we would have to segment it into several helper functions in order to reduce the cognitive complexity for the Milestone 3.

app/src/main/java/com/android/streetworkapp/ui/navigation/TopAppBar.kt
```diff
@@ -0,0 +1,37 @@
...
+import androidx.compose.material3.TopAppBar
+import androidx.compose.material3.TopAppBarDefaults
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun TopAppBarWrapper(navigationActions: NavigationActions, topAppBarManager: TopAppBarManager?) {
+  topAppBarManager?.let {
```
tercierp, line 17: Improvement: instead of the nested block you could just add this line and remove the block
```kotlin
if (topAppBarManager == null) return
TopAppBar(...)
```

app/src/main/java/com/android/streetworkapp/ui/navigation/TopAppBarManager.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.compose.ui.graphics.vector.ImageVector
+
+class TopAppBarManager(
+    private var title: String = "",
+    private var hasNavigationIcon: Boolean = false,
+    private var navigationIcon: ImageVector? = null
+) {
+
+  companion object {
+    val DEFAULT_TOP_APP_BAR_NAVIGATION_ICON = Icons.AutoMirrored.Filled.ArrowBack
```
tercierp, line 14: Suggestion: change the `val` into `var` this improves immutability and reduces potential side effects.

app/src/main/java/com/android/streetworkapp/ui/navigation/TopAppBarManager.kt
```diff
@@ -0,0 +1,35 @@
...
+  }
+  /** Changes the TopAppBar title */
+  fun setTopAppBarTitle(newTitle: String) {
+    this.title = newTitle
+  }
+
+  /** Get the TopAppBar title */
+  fun getTopAppBarTitle(): String {
+    return this.title
+  }
```
tercierp, line 24: Suggestion: You could change this getter with a more Kotlin way to get the `title` 

```kotlin
val topAppBarTitle: String
        get() = title
```       
This Kotlin "getter" could be used like this:
```kotlin
val topAppBarManager = //init your top bar manager
val title = topAppBarManager.topAppBarTitle
```

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -23,66 +24,56 @@ import androidx.compose.ui.unit.sp
...
-                  modifier = Modifier.size(220.dp, 50.dp).testTag("profileTrainButton")) {
-                    Text(text = "Train with a friend", fontSize = 17.sp)
-                  }
-            }
-      })
+          Spacer(modifier = Modifier.height(10.dp))
+          // button to train with a friend
+          Button(
+              onClick = {},
+              colors = ButtonDefaults.buttonColors(Color(0xFFA53A36)),
```
tercierp, line 73: Improvement: probably not in your task range but we should avoid setting directly a color instead use the palette of the `Color.kt`

tercierp (changes requested):
Great work on this refactoring! The changes have notably improved the code’s readability and maintainability. I found a few additional improvements and suggestions.
I also have a question about the paddings, I imagine there's a need to specify padding, but I don't understand in what situations it should be specified.

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
Alex720p: you're supposed to pass it as a param when using scaffold. Haven't tested personnally but a non zero value should be passed when bottom bar is set, otherwise some content could be hidden

app/src/main/java/com/android/streetworkapp/ui/navigation/TopAppBar.kt
Alex720p: I think it's prettier like this :), I did change `topAppBarManager` to `it` inside the block, was pretty stupid on my part for rereferencing it inside the block

app/src/main/java/com/android/streetworkapp/ui/navigation/TopAppBarManager.kt
Alex720p: I'm not sure I understand this, changing the val into var would be worse for immutability

app/src/main/java/com/android/streetworkapp/ui/navigation/NavigationActions.kt
```diff
@@ -3,6 +3,8 @@ package com.android.streetworkapp.ui.navigation
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.AccountCircle
 import androidx.compose.material.icons.outlined.Place
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.setValue
```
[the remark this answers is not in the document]
Alex720p: done

app/src/androidTest/java/com/android/streetworkapp/ui/navigation/TopAppBarTest.kt
Alex720p: noted

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Alex720p: Since we'll have to rework the Profile screen anyways, it might be a good idea to do it then

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -98,63 +96,45 @@ fun AddEventScreen(
...
+            Text(
+                text = "How many participants do you want?",
+                fontSize = 18.sp,
+            )
+            ParticipantNumberSelection(event)
           }
-        }
+      FloatingActionButton(
+          onClick = {
+            // TODO: check that it works on the database
```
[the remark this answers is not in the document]
Alex720p: done

app/src/main/java/com/android/streetworkapp/ui/navigation/TopAppBarManager.kt
tercierp: No problem, resolved

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
tercierp: Yeah I agree

app/src/main/java/com/android/streetworkapp/ui/navigation/TopAppBarManager.kt
tercierp: Sorry I was indeed confused, my bad resolved

tercierp (approved):
Thank you for the changes, good job !
I let you decide if you want one more review or directly merge

PR #59, opened 2024-11-02
Hotfix: Update the add event screen to have a park id in each event.

alvaro1080:
**Fixes in the AddEvent screen**
This pull request was created to update and fix the screen used to add a new event.

**Changes**
- Each new event has now a park id linked to it
- The user can not now chose a date in the past for an event (Note: a toDate() function was used, whose precision is not perfect. It does work to prevent the user from choosing a date that is yesterday or in 1970, but the user can still chose a time that is a few second before the current time. This could be improved but it already work as a sanity check)
- Update the UI of addEvent to be coherent with the changes made in the scaffold

### DIFF
--- a/app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
+++ b/app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
@@ -133,7 +133,7 @@ class AddEventTest {
     composeTestRule.onNodeWithTag("addEventButton").performClick()
     composeTestRule
         .onNodeWithTag("addEventScreen")
-        .assertIsDisplayed() // we that that a click with no changes in the default value does not
+        .assertIsDisplayed() // we want that a click with no changes in the default value does not
     // make the user leave the screen
   }
 
@@ -154,7 +154,7 @@ class AddEventTest {
     composeTestRule.onNodeWithTag("addEventButton").performClick()
     composeTestRule
         .onNodeWithTag("addEventScreen")
-        .assertIsDisplayed() // we that that a click with changes in the time but no changes in the
+        .assertIsDisplayed() // we want that a click with changes in the time but no changes in the
     // default value of the title does not make the user leave the screen
   }
 }
--- a/app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
+++ b/app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
@@ -134,7 +134,7 @@ class AddEventTest {
     composeTestRule
         .onNodeWithTag("addEventScreen")
         .assertIsDisplayed() // we that that a click with no changes in the default value does not
-                             // make the user leave the screen
+    // make the user leave the screen
   }
 
   @Test
@@ -155,6 +155,6 @@ class AddEventTest {
     composeTestRule
         .onNodeWithTag("addEventScreen")
         .assertIsDisplayed() // we that that a click with changes in the time but no changes in the
-                             // default value of the title does not make the user leave the screen
+    // default value of the title does not make the user leave the screen
   }
 }
--- a/app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
+++ b/app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
@@ -121,4 +121,40 @@ class AddEventTest {
     composeTestRule.onNodeWithTag("addEventScreen").assertIsDisplayed()
     composeTestRule.onNodeWithTag("addEventButton").assertIsDisplayed()
   }
+
+  @Test
+  fun addEventScreenWithUnchangedTimeDoesNotChangeScreen() {
+    `when`(eventViewModel.getNewEid()).thenReturn("test")
+    composeTestRule.setContent {
+      AddEventScreen(navigationActions, parkViewModel, eventViewModel, userViewModel)
+    }
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -104,6 +107,7 @@ fun AddEventScreen(
           modifier = Modifier.fillMaxWidth(),
           verticalArrangement = Arrangement.spacedBy(18.dp),
           horizontalAlignment = Alignment.CenterHorizontally) {
+            Spacer(modifier = Modifier.size(24.dp))
```
SaturneV, line 110: I recommend that you add in the PR description that you've modified some UI elements, so that the description is up to date.

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -115,22 +119,31 @@ fun AddEventScreen(
           }
       FloatingActionButton(
           onClick = {
-            if (event.title.isEmpty()) {
+            if (event.date.toDate() < Calendar.getInstance().time) {
```
SaturneV, line 122: To avoid accepting events in the past, we could authorize events only from the **present moment + 5 minutes**. Because I think that creating an event that takes place at the precise moment of its creation doesn't make sense, and so shouldn't be allowed. What do you think about this?

app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
```diff
@@ -121,4 +121,40 @@ class AddEventTest {
     composeTestRule.onNodeWithTag("addEventScreen").assertIsDisplayed()
     composeTestRule.onNodeWithTag("addEventButton").assertIsDisplayed()
   }
+
+  @Test
+  fun addEventScreenWithUnchangedTimeDoesNotChangeScreen() {
```
SaturneV, line 126: Is it easy for you to add a test that tests when the event title is empty (check the actual SonarCloud report)? With this, you could expect 100% line coverage, but if it's too complex, don't worry, the coverage is already quite high.

SaturneV (changes requested):
It's a great PR, good job! I'll let you look at my comments and respond to them, especially adding in the PR description the UI changes you've made.

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
alvaro1080: It is done now, thank you!

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
alvaro1080: Since the toDate() fonction makes the timeStamp of event.date loose in precision, I am not sure that this addition of 5 minutes would be really useful. For the moment this fix works for obvious mistakes of the user (for instance if he chooses a day before today or a really old date). It is also a fix for the "by default date" bug that we encountered in the last meeting: the date was by default 1970. I think that not allowing an event a the present moment makes sens and we should talk about it in the next meeting. For the scope of this PR, I think the current solution is the most relevant.

app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
alvaro1080: I agree with you. This is the point of the second test, but SonarCloud does not agree. (On the jaccoco report in the other hand, the lines are considered covered, I am a bit lost with this). Since the coverage is high enough in the whole project and in this PR, I think it is better to stay on this coverage.

misterM125 (approved):
Very good code overall and everything seems to be working properly as the sanity checks are done and parkId is properly added to the database
I have a few minor syntax corrections that could improve ui and code readability but i will let you change it at your leisure

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -115,22 +119,31 @@ fun AddEventScreen(
...
+              Toast.makeText(
+                      context, "Please select a valid time in the future", Toast.LENGTH_SHORT)
+                  .show()
+            } else if (event.title.isEmpty()) {
               Toast.makeText(context, "Please fill the title of the event", Toast.LENGTH_SHORT)
                   .show()
             } else {
               eventViewModel.addEvent(event)
+              parkViewModel.currentPark.value?.let {
+                parkViewModel.addEventToPark(it.pid, event.eid)
```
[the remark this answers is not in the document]
alvaro1080: You are right, but this value might be null. But you are right, I will update this line and use the event.parkId, since in the worst case it has the default parkId value of a park. Thank you for your suggestion!

app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/AddEventTest.kt
SaturneV: No problem then, thank you for the clarification!

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
SaturneV: Thank you for clarifying the problem! Then we'll talk about it again at a future meeting, in which we'll also clarify the default date.

SaturneV (approved):
From my side, all my questions and suggestions have been answered, so for me this PR is ready for merge. Good job!

PR #56, opened 2024-10-31
Feat/Integrate Park Overview in Map UI

SaturneV:
### Summary

This pull request integrates the Park MVVM into the Map UI. When a park is selected, it is fetched from the database or created if it does not exist, and its information is displayed in the park overview screen.

### Changes

- Modified `MainActivity` to integrate the park view model.
- Modified `MapUI` to integrate the park view model.
- Modified `ParkOverview` to integrate the park view model.

### Testing

IN PROGRESS

PR #63, opened 2024-11-03
Feat/Fetching function in Park MVVM

SaturneV:
### Summary

This pull request adds a new function to the Park MVVM that fetches a park by location ID if it exists or creates a new park if it does not. This enhancement streamlines the process of managing parks based on their location IDs.

### Changes

- Added the function to the `ParkRepository` interface.
- Implemented the function in the `ParkRepositoryFirestore` class.
- Implemented the function in the `ParkViewModel` class.

### Testing 

- Added new tests to `ParkRepositoryFirestore` to test the function
- Added new tests to `ParkViewModel` to test the function
- Implemented `ParkUtils` tests to test the creation of a default park

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/model/park/ParkUtils.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/park/ParkUtils.kt
@@ -6,15 +6,13 @@ import com.android.streetworkapp.model.parklocation.ParkLocation
  * Create a new default park given the park ID and location
  *
  * @param pid The park ID.
- * @param lon The park longitude.
- * @param lat The park latitude.
- * @param locationId The park location ID.
+ * @param parkLocation The park location.
  */
-fun createDefaultPark(pid: String, lon: Double, lat: Double, locationId: String): Park {
+fun createDefaultPark(pid: String, parkLocation: ParkLocation): Park {
   return Park(
       pid = pid,
-      name = "Default Park $locationId",
-      location = ParkLocation(lon, lat, locationId),
+      name = "Default Park ${parkLocation.id}",
+      location = parkLocation,
       imageReference = "",
       rating = 1f,
       nbrRating = 0,
--- a/app/src/main/java/com/android/streetworkapp/model/park/ParkUtils.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/park/ParkUtils.kt
@@ -1,0 +1,24 @@
+package com.android.streetworkapp.model.park
+
+import com.android.streetworkapp.model.parklocation.ParkLocation
+
+/**
+ * Create a new default park given the park ID and location
+ *
+ * @param pid The park ID.
+ * @param lon The park longitude.
+ * @param lat The park latitude.
+ * @param locationId The park location ID.
+ */
+fun createDefaultPark(pid: String, lon: Double, lat: Double, locationId: String): Park {
+  return Park(
+      pid = pid,
+      name = "Default Park $locationId",
+      location = ParkLocation(lon, lat, locationId),
+      imageReference = "",
+      rating = 1f,
+      nbrRating = 0,
+      capacity = 1,
+      occupancy = 0,
+      events = emptyList())
+}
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
```diff
@@ -80,6 +81,18 @@ open class ParkViewModel(private val repository: ParkRepository) : ViewModel() {
...
+  /**
+   * Get or create a park by its location.
+   *
+   * @param location The park location.
+   */
+  fun getOrCreateParkByLocation(location: ParkLocation) {
+    viewModelScope.launch {
+      val park = repository.getOrCreateParkByLocation(location)
+      _park.value = park
+    }
```
alvaro1080, line 93: Comment: after some test on my side, this function work as expected. There is unfortunately a bug on the first time that this function is used if there is no park already on the database: The overview screen will show the n-1 park, and not the current n park. To resolve this I created a function :

fun setDefaultPark() {
    getOrCreateParkByLocation(ParkLocation(0.0,0.0,"default"))
  }

in ParkViewModel that is used at the launch of the MapUI. This solves the bug and the map works as intended. You might find it useful for the implementation in the MapUi. You can already define the setDefaultPark() function in this PR (that is my advice) or solve this problem in the next PR.

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
```diff
@@ -80,6 +81,18 @@ open class ParkViewModel(private val repository: ParkRepository) : ViewModel() {
...
 
+  /**
+   * Get or create a park by its location.
+   *
+   * @param location The park location.
+   */
+  fun getOrCreateParkByLocation(location: ParkLocation) {
+    viewModelScope.launch {
+      val park = repository.getOrCreateParkByLocation(location)
+      _park.value = park
```
alvaro1080, line 92: Comment: I think that having a "Park" and a "CurrentPark" is more confusing than useful. I see why this is implemented in order to distinguish the 2 uses, but we have, at the moment, not really used 2 separate versions. My advice would be to keep only one but I see why you would prefer to keep the 2 separate.

alvaro1080 (changes requested):
Thank you for your Pull Request! The function works on my side, except for a bug that is detailed on the next comment. The code is clean and will be really useful to connect the events and the parks. I also have a few comments that I would like to discuss:

app/src/main/java/com/android/streetworkapp/model/park/ParkUtils.kt
```diff
@@ -0,0 +1,22 @@
+package com.android.streetworkapp.model.park
+
+import com.android.streetworkapp.model.parklocation.ParkLocation
```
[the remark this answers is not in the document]
SaturneV: I believe this file fits better in the `park` package, as it groups feature-specific code together. This function won’t be used in other parts of the project — it serves only as a formatter for creating parks in the database, then is only used in the ViewModel. Do you agree?

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
SaturneV: Basically I agree with you - right now in our project we only need a current park. However, I'd rather keep potential features for future implementation than remove something already implemented. What do you think of it?

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
SaturneV: After testing on my side, I observed the same problem. But since the viewmodel itself is functional and correctly working, I think it's up to the integration to adapt to the viewmodel. So I'm leaving this modification to [this](https://github.com/SwEnt-Group8/Street-work-app/pull/56) PR, does that work for you?

alvaro1080: I think this is a good way to handle this problem. Thank you!

app/src/main/java/com/android/streetworkapp/model/park/ParkUtils.kt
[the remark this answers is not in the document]
alvaro1080: Your argument convinced me. I agree.

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
alvaro1080: I understand your argument. I think we can keep it as it is for the moment.

alvaro1080 (approved):
My suggestions are now clarified in the comments. Thank you for your answers!

PR #67, opened 2024-11-04
Feat/friend screen in profile screen

Simmanz:
## Friend screen in profile

Profile-with-basic-friend-screen

Profile-with-basic-friend-screen-empty

### Profile UI

- Now displays a **list of friends** (hardcoded list to cirumvent MVVM issues).
- If **no friends** (empty list), then the display is replaced with a **text**.
- Also now display current user (placeholder) **score** and **username**.
- Removed **QR-code** placeholder because we will use **NFC** or **UID-adding**.
- Clicking on the **train with a friend** button will now display a toast saying that it is not implemented yet.

### Components to display friends

- `DisplayFriend()` to display one element in the friend list
- `DisplayFriendList()` to display a list of friends (User).

Currently displayed friend information is :
- The username
- The score
- A placeholder profile picture

### Tests

- `ProfileScreenTest()` has been **updated**.
- `ProfileFriendListTest()` verifies that the components to display friends are working properly.

### To test the feature

Simply run the `mainActivity` and go to the **profile screen**.
You can uncomment the following line : `friendList = emptyList()` in order to verify the no friends behavior

### Improvements for future tasks

- Make the friend screen UI looks like the Figma.
- MVVM linking works (fetching currentUser and friends).


app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -33,47 +41,130 @@ fun ProfileScreen(
...
+    Image(
+        painter = painterResource(id = R.drawable.profile),
+        contentDescription = "profile picture",
+        modifier = Modifier.size(75.dp).testTag("friendProfilePicture"))
+
+    Spacer(modifier = Modifier.width(10.dp))
+
+    Column(modifier = Modifier.fillMaxWidth().padding(padding)) {
+
+      // Small spacing between elements
```
alvaro1080, line 168: Refactor: I do not obtain the same result as you in the UI. The spacing (at least in my laptop) between the friends are way too large. Here is a version that gives a more convinient spacing between friends:

(in summary, you should remove all the paddings and remove some of the spacers)

```
@Composable
fun DisplayFriendList(friends: List<User>, padding: PaddingValues = PaddingValues(0.dp)) {
  return if (friends.isNotEmpty()) {
    // LazyColumn is scrollable
    LazyColumn(
        contentPadding = PaddingValues(vertical = 0.dp),
        modifier =
            Modifier.fillMaxWidth()
                .testTag("friendList")) {
          items(friends) { friend -> DisplayFriend(friend) }
        }
  } else {
    Text(
        modifier = Modifier.padding(padding).testTag("emptyFriendListText"),
        fontSize = 20.sp,
        text = "You have no friends yet :(")
  }
}

```

```
@Composable
fun DisplayFriend(friend: User) {
  Row(modifier = Modifier.fillMaxWidth().testTag("friendRow")) {
    // profile placeholder
    Image(
        painter = painterResource(id = R.drawable.profile),
        contentDescription = "profile picture",
        modifier = Modifier.size(75.dp).testTag("friendProfilePicture"))

    Column(modifier = Modifier.fillMaxWidth()) {

      Text(fontSize = 22.sp, text = friend.username, modifier = Modifier.testTag("friendUsername"))

      // score
      Text(
          text = "Score: ${friend.score}",
          fontSize = 22.sp,
          modifier = Modifier.testTag("friendScore"))
    }
  }
}


```


Please note that those are suggestion based on the current code. This could be improved in other way if you wish, but it is important to change the way friends are displayed

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -33,47 +41,130 @@ fun ProfileScreen(
...
-              modifier = Modifier.size(220.dp, 50.dp).testTag("profileTrainButton")) {
-                Text(text = "Train with a friend", fontSize = 17.sp)
+                  Spacer(modifier = Modifier.height(10.dp).testTag("profileInfoSpacer3"))
+
+                  // button to train with a friend
+                  Button(
+                      onClick = {
+                        Toast.makeText(context, "Not implemented yet", Toast.LENGTH_LONG).show()
+                      },
+                      colors = ButtonDefaults.buttonColors(Color(0xFFA53A36)),
```
alvaro1080, line 108: Improvement: the color could be defined in another file as a constant

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -33,47 +41,130 @@ fun ProfileScreen(
...
+  // Fake data extraction from the MVVM :
+  val alice =
+      User("uid_Alice", "Alice", "alice@gmail.com", 42, listOf("Z8qgzNOoQOR09x6QQYahcyzVXfE2"))
+
+  val bob = User("uid_Bob", "Bob", "bob@gmail.com", 64, emptyList())
+
+  val currentUser =
+      User("uid_current", "Current User", "user@gmail.com", 42424, listOf(alice.uid, bob.uid))
+
+  var friendList = listOf(alice, bob)
```
alvaro1080, line 56: Improvement: even if it is meant to be removed later, we should use a val here instead of a var

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -33,47 +41,130 @@ fun ProfileScreen(
...
+  // friendList = emptyList()
```
alvaro1080, line 57: Suggestion: you can either add a val at the start of this commented line, or just remove it

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -33,47 +41,130 @@ fun ProfileScreen(
...
         modifier = Modifier.fillMaxSize().padding(innerPaddingValues).testTag("ProfileColumn"),
         horizontalAlignment = Alignment.CenterHorizontally,
         verticalArrangement = Arrangement.spacedBy(10.dp, Alignment.CenterVertically)) {
-          // profile placeholder
 
-          Image(
-              painter = painterResource(id = R.drawable.profile),
-              contentDescription = "profile picture",
-              modifier = Modifier.size(200.dp))
+          Spacer(modifier = Modifier.height(4.dp).testTag("profileSpacer"))
```
alvaro1080, line 64: Suggestion: I am not sure that the test tags for spacers are relevant, since testing spacers seem unnecessary(appart from testing that they are displayed, but it might be too much). You could remove them.

alvaro1080, line 64: Question: I am not sure why this spacer in particular is used? Is the space above the picture profile not enough?

alvaro1080 (changes requested):
Thank you for you Pull Request! Since we changed the way friends are added, the QR was not relevant anymore, this PR is thus useful to transition to the NFC based friend system. 
I have not the same UI as the one shown in the PR. This is the main point of my review that is in one of the comment bellow.
I also have a few suggestions that I would like to discuss:

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: I removed it. It was for you to be able to see both screens easily.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: Because we were usually testing each composable component I choose to verify the existence of the spacers too.

Simmanz: Before the scaffold merge it was needed to space things more specifically than simply using padding. I removed it.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: I am unsure what kind of device you use. I removed all padding and kept only one spacer which prevents the friend username from being too high relatively to the image. Is it better like this for you ?

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: I also removed the others spacers and instead put small padding to prevent things from breaking on other formats

alvaro1080 (approved):
The Ui now works as describe in the scope of this PR. Thank you for your updates!

PR #69, opened 2024-11-05
Feat/add user profile picture

kzepfl:
## Changes

- Changed the image in `Profile.kt` to display the current user profile picture
- Added a dependecies in `build.gradle.kts` for coil (needed to display the profile picture)

## Testing

- Added new test to verify the picture is correctly displayed (`ProfileScreenTest.kt`)

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
+++ b/app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
@@ -2,6 +2,7 @@ package com.android.streetworkapp.ui.profile
 
 import android.widget.Toast
 import androidx.compose.foundation.Image
+import androidx.compose.foundation.border
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -65,12 +75,27 @@ fun ProfileScreen(
...
-                // profile placeholder
-                Image(
-                    painter = painterResource(id = R.drawable.profile),
-                    contentDescription = "profile picture",
-                    modifier = Modifier.size(200.dp).testTag("profilePicture"),
-                )
+                // fetch profile picture from firebase
+                val auth = Firebase.auth
+                val user = auth.currentUser
+                val photo = user?.photoUrl
```
SaturneV, line 81: Isn't it is more compact to juste have it in one line like that: 
`val photo = Firebase.auth.currentUser?.photoUrl`
You don't use the intermediate variables anywhere.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -65,12 +75,27 @@ fun ProfileScreen(
...
+                val placeholderUri =
+                    "https://picsum.photos/200/300".toUri() // Placeholder image URI
```
SaturneV, line 83: In my opinion, you should rather store this profile picture locally - in the case of the app not connected to internet, we should be able to display it.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -65,12 +75,27 @@ fun ProfileScreen(
...
+
+                // display the profile picture
+                AsyncImage(
+                    model =
+                        ImageRequest.Builder(LocalContext.current)
+                            .data(photo ?: placeholderUri)
+                            .placeholder(R.drawable.profile)
+                            .build(),
+                    contentDescription = null,
```
SaturneV, line 92: Please think about adding a `contentDescription`, it improve the app accesibility.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -65,12 +75,27 @@ fun ProfileScreen(
...
 
-                // profile placeholder
-                Image(
-                    painter = painterResource(id = R.drawable.profile),
-                    contentDescription = "profile picture",
-                    modifier = Modifier.size(200.dp).testTag("profilePicture"),
-                )
+                // fetch profile picture from firebase
+                val auth = Firebase.auth
+                val user = auth.currentUser
```
SaturneV, line 80: From a more general point of view, this UI file will fetch the current user, and that user will be used across all fields in the UI. Wouldn’t it be better to fetch the current user outside of the `Box` element, rather than within this very specific part of the file?

SaturneV (changes requested):
Good PR and great work! I’ll let you take a look at my comments. I have a general suggestion - it would be a good idea to include a screenshot of the modified UI element to enhance understanding when reading the PR description. Don't forget to add labels to the PR too.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
kzepfl: added

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
kzepfl: I removed the placeholder URI

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
kzepfl: fixed

SaturneV: I was more thinking of first fetching the users and then fetching the photo specifically. The user will be used in several parts of the file so it would be nice to have a global `user` variable that we can reuse to access `username`, `score`, etc.

kzepfl: I dont think we need the firebase user, it is only used to fetch the gmail profile picture.
The nb of point and other feature are inside the user MVVM

SaturneV: Indeed, I hadn't thought of that! In this case it's all good.

PR #76, opened 2024-11-07
Feat/Event actualisation in event and park overview

misterM125:
test


alvaro1080 (changes requested):
Thank you for this Pull Request! The functionality of this new feature seem to work on my side: clicking on a park displays the correct park with the correct events. Congratulations! 
I have a few comments that I would like to discuss:

alvaro1080 (approved):
Thank you for the updates!

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -66,4 +67,21 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+          title = document["title"] as String,
+          description = document["description"] as String,
+          participants = (document["participants"] as Long).toInt(),
+          maxParticipants = (document["maxParticipants"] as Long).toInt(),
+          date = document["date"] as Timestamp,
+          owner = document["owner"] as String,
+          listParticipants = document["listParticipants"] as List<String>,
+          parkId = document["parkId"] as String)
+    } else {
+      null
```
tercierp, line 84: Suggestion: would be nice to add a `Log.e` in order to track some unexpected behaviours

app/src/main/java/com/android/streetworkapp/model/event/EventViewModel.kt
```diff
@@ -24,16 +45,21 @@ open class EventViewModel(private val repository: EventRepository) : ViewModel()
...
+  fun getEvents(park: Park) {
+    viewModelScope.launch {
+      repository.getEvents(
+          park,
+          onSuccess = {
+            if (it.isEmpty()) {
+              Log.d("EventViewModel", "No events found")
+              setUiState(EventOverviewUiState.Empty)
+            } else {
+              Log.d("EventViewModel", "lesgo it is not empty event: $it")
```
tercierp, line 57: Note: nice debug logcat 😂

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
```diff
@@ -57,7 +57,7 @@ open class ParkViewModel(private val repository: ParkRepository) : ViewModel() {
   fun getParkByPid(pid: String) {
     viewModelScope.launch {
       val fetchedPark = repository.getParkByPid(pid)
-      _park.value = fetchedPark
+      _currentPark.value = fetchedPark
```
tercierp, line 60: Question: I was not there when @SaturneV designed the MVVM and I guess he must have explained the difference between a `park` and a `currentPark`. 
With this change I'm a bit lost, I see only two possibilities: 

1. Either the `currentPark` and the `park` make sense as in MVVM (which means that there are cases where you need to have both at the same time), or the `currentPark` and the `park` make sense as in MVVM (which means that there are cases where you need to have both at the same time). In this case, certain functions need to be added to make `currentPark` behave like `park`.
2. Or delete `park` and work only with `currentPark` in this MVVM (because you never need both at the same time).

Could you enlighten me on this question?

app/src/test/java/com/android/streetworkapp/model/park/ParkViewModelTest.kt
```diff
@@ -192,7 +192,7 @@ class ParkViewModelTest {
     parkViewModel.getParkByPid("123")
     testDispatcher.scheduler.advanceUntilIdle()
 
-    val observedPark = parkViewModel.park.first()
+    val observedPark = parkViewModel.currentPark.first()
```
tercierp, line 195: Question: same confusion problem is there a difference between a currentPark and a park? If so, when is it useful?

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -66,4 +67,21 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+    return if (document.exists()) {
+      Event(
+          eid = document.id,
+          title = document["title"] as String,
+          description = document["description"] as String,
+          participants = (document["participants"] as Long).toInt(),
+          maxParticipants = (document["maxParticipants"] as Long).toInt(),
+          date = document["date"] as Timestamp,
+          owner = document["owner"] as String,
+          listParticipants = document["listParticipants"] as List<String>,
```
tercierp, line 81: Question: I'm not sure, but maybe there's a way to remove this unchecked cast ?

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
```diff
@@ -40,33 +40,43 @@ import androidx.compose.ui.res.painterResource
...
+
   Box(modifier = Modifier.padding(innerPadding).fillMaxSize().testTag("parkOverviewScreen")) {
     Column {
-      ImageTitle(image = null, title = park.name) // TODO: Fetch image from Firestore storage
-      ParkDetails(park = park)
-      EventItemList(eventViewModel) // TODO: Fetch events from Firestore
+      ImageTitle(image = null, title = currentPark.value?.name ?: "loading...")
+      // TODO: Fetch image from Firestore storage
+      currentPark.value?.let { ParkDetails(park = it) }
+      EventItemList(eventViewModel, navigationActions) // TODO: Fetch events from Firestore
```
tercierp, line 79: Question: why this is still a TODO since on my phone I do see the `events` when I pressed on a `park`?

tercierp (changes requested):
Good work, I've asked you a number of questions that could potentially lead to changes, so I'll let you get back to me and make the necessary changes according to your answer.

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
misterM125: I agree, good call!

app/src/main/java/com/android/streetworkapp/model/event/EventViewModel.kt
misterM125: oops...

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
misterM125: The reason for this change is mainly because of functionality issues with the park overview:
Essentially i need to update two different values to make it work properly.
If you look at the map ui, the `getOrCreatePark()` function is called when clicked on a marker to set the `Park` value and then the `getParkByPid()` function is called in the park overview. 
It is necessary to set the current park in the park overview for two reasons:
- first is because the park overview could be accessed in other ways than the map, for  example you return to the park overview after creating an event. If `getParkByPid()` wasn't called at this point, events would not be updated a new one has been added to firebase
- The second reason is for consistency issues with the park overview: if the current park is set to different values in values in quick succesion (i.e in the map ui and park overview) there seems to be weird asynchronous behavior where the values might not be set in the right order and events might not be displayed properly. This is the reason i am using both values for my park overview and why i changed the set value of `getParkByPid`

I acknowledge that it might not entirely be clear when looking at the code but this is the best way i have found to solve these issues. Let me know if you think there is a better way to resolve this problem

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
misterM125: None honestly, i'll remove it

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
tercierp: I'm not sure I fully understood your answer so I'll try to rephrase it. And then I'll show you what bothers me about your answer.

You need currentpark and park, for example in mapUI where you have several parks to display.
But what I still don't understand is why modify the `getParkByPid` function the way you did.
I don't think you've seen it, but in the MVVM that @SaturneV has implemented, the `loadCurrentPark` function does exactly what you want it to do! 

To be more formal, I'll go into my problem description.
In your PR you change the `getParkByPid` function:
1. Before your PR
```kotlin
fun getParkByPid(pid: String) {
    viewModelScope.launch {
      val fetchedPark = repository.getParkByPid(pid)
      _park.value = fetchedPark
      }
  }
```

2. After your PR
```kotlin
fun getParkByPid(pid: String) {
    viewModelScope.launch {
      val fetchedPark = repository.getParkByPid(pid)
      _currentPark.value = fetchedPark
    }
  }
```

3. But `loadCurrentPark` is the exact same function as your modified `getParkByPid` function.
```kotlin
fun loadCurrentPark(pid: String) {
    viewModelScope.launch {
      val fetchedPark = repository.getParkByPid(pid)
      _currentPark.value = fetchedPark
    }
  }
```
There's no world in which 2 and 3 can coexist in the same MVVM, do you agree ?

My suggestion is that you use `loadCurrentPark` where you used to use `getParkByPid` and go back to the version before the changes to the `getParkByPid` function.

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
tercierp: Resolved

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
tercierp: Yep well done

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
misterM125: you are right, i did not realise that there already was a function doing the exact same thing, I have applied the suggestion
Thank you for pointing that out

tercierp (approved):
Thanks for the changes you've made, it's approved.

PR #77, opened 2024-11-07
Feat/impove profile screen

Simmanz:
### Friend list

- New format overall
- Now has **separating lines** between friend elements
- Now displays a **placeholder status** for friends

### Profile UI

- Changed the **whole structure**
- **Removed** "Train with a friend" button as it is no longer relevant
- Moved the add friend button and other text elements

### Tests

Adapted both tests files because some elements were deleted / added and others were slightly modified.

### Before merging

Will wait for the final user MVVM hotfix to merge it inside before merging in main (in order not to break all the changes done to the UI).

## Merge with the user MVVM fix

- Kept my changes (new UI, deletion of some tests)
- Kept hotfix changes (logic to handle current user and friend list as mutableState, new tests, removal of old hardcoded tests)

### New components

Code to display score and username was starting to get heavy, made a separated functions for each of them : 

- `DisplayUsername()`
- `DisplayScore()`
- Refactored test file `ProfileFriendListTest` into `ProfileComponentsTest` and created new related tests.

## Future updates for later

- Implement settings button on the top right (topAppBar => actions section could be the solution)
- Implement settings button for each friend element (if we decide to keep them later)
- Available - The friend elements can be added the `.clickable()` property => way to order the friends (drag & drop ?)


app/src/androidTest/java/com/android/streetworkapp/ui/profile/ProfileComponentsTest.kt
```diff
@@ -36,6 +66,14 @@ class ProfileFriendsListTest {
         .assertExists()
         .assertIsDisplayed()
         .assertTextEquals("Score: ${friend.score}")
+
+    composeTestRule
+        .onNodeWithTag("friendStatus")
+        .assertExists()
+        .assertIsDisplayed()
+        .assertTextEquals("Definitely not a bot")
```
tercierp, line 74: Improvement: Given that not everyone has the figma nearby, I think that hardcoding funny values without specifying what they are will confuse some teammates

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -148,20 +152,17 @@ fun ProfileScreen(
...
-    // LazyColumn is scrollable
-    LazyColumn(
-        contentPadding = PaddingValues(vertical = 0.dp),
-        modifier = Modifier.fillMaxWidth().padding(horizontal = 16.dp).testTag("friendList")) {
-          items(friends) { friend ->
-            if (friend != null) {
-              DisplayFriend(friend)
-            }
-          }
+    LazyColumn(modifier = Modifier.fillMaxSize().testTag("friendList")) {
```
tercierp, line 155: Note: Good job using `LazyColumn` for displaying friend items in `DisplayFriendList`. Lazy loading is beneficial for handling a potentially large list of friends.

tercierp (changes requested):
One minor improvement requested in the test and the `Profile.kt`, otherwise very good work !

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -173,28 +174,50 @@ fun DisplayFriendList(friends: List<User?>) {
...
+              fontSize = 18.sp,
+              fontWeight = FontWeight.Bold,
+              modifier = Modifier.testTag("friendUsername"))
+          Text(
+              text = "Score: ${friend.score}",
+              fontSize = 14.sp,
+              color = Color.Gray,
+              modifier = Modifier.testTag("friendScore"))
+          Text(
+              text = "Definitely not a bot",
```
tercierp, line 206: Improvement: same as the improvement in the tests

app/src/androidTest/java/com/android/streetworkapp/ui/profile/ProfileComponentsTest.kt
Simmanz: I made default values explicit using constants (upper case vals), is that good for you ?

tercierp (approved):
Thanks for your changes !

PR #79, opened 2024-11-07
Hotfix remove the `Livedata`

tercierp:
This PR aims to delimit all `Livedata` in the `User` MVVM (as explained in the issue after a discussion with Simon we decided to delete `Livedata`).


Simmanz (approved):
The user MVVM finally works ! User can correctly add friends and see them updated in the friend list in the profile screen, I tried manually and it works :

friendlist-final-test

PR #84, opened 2024-11-08
Feat/rating for `Park`

tercierp:
In this branch, park data has been modified to include the rating, 

The idea behind the implementation is to minimize calculation time, bearing in mind that we are limited by Firestore 

```kotlin
data class Park(
    val pid: String,
    var name: String,
    var location: ParkLocation,
    var imageReference: String,
    var rating: Float = 0f,
    var nbrRating: Int = 0,
    var capacity: Int,
    var occupancy: Int,
    var events: List<String>,
    var votersUIDs: List<String> = emptyList()
)
```
If a `User` add a new rating we update the rating with the following formula:

$$updatedRating = \frac{newRating + nbrRating \cdot rating}{nbrRating+1}$$

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
@@ -135,32 +135,6 @@ class ParkRepositoryFirestore(private val db: FirebaseFirestore) : ParkRepositor
   }
 
   /**
-   * Add a rating to a park.
-   *
-   * @param pid The park ID.
-   * @param rating The rating to add from 1 to 5.
-   */
-  override suspend fun addRating(pid: String, rating: Int) {
-    require(pid.isNotEmpty()) { PID_EMPTY }
-    require(rating in 1..5) { INVALID_RATING_MESSAGE }
-    try {
-      val document = db.collection(COLLECTION_PATH).document(pid).get().await()
-      val currentRating = document.getDouble("rating") ?: 0.0
-      val currentNbrRating = document.getLong("nbrRating")?.toInt() ?: 0
-
-      val newNbrRating = currentNbrRating + 1
-      val newRating = ((currentRating * currentNbrRating) + rating) / newNbrRating
-
-      db.collection(COLLECTION_PATH)
-          .document(pid)
-          .update(mapOf("rating" to newRating, "nbrRating" to newNbrRating))
-          .await()
-    } catch (e: Exception) {
-      Log.e("FirestoreError", "Error adding rating to park: ${e.message}")
-    }
-  }
-
-  /**
    * Delete a rating from a park.
    *
    * @param pid The park ID.
--- a/app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
@@ -295,6 +295,47 @@ class ParkRepositoryFirestore(private val db: FirebaseFirestore) : ParkRepositor
   }
 
   /**
+   * Add a rating to a park if the user has not already rated it.
+   *
+   * @param pid The park ID.
+   * @param uid The user ID of the person rating.
+   * @param rating The rating to add from 1 to 5.
+   */
+  override suspend fun addRating(pid: String, uid: String, rating: Int) {
+    require(pid.isNotEmpty()) { PID_EMPTY }
+    require(rating in 1..5) { "Rating must be between 1 and 5." }
+
+    try {
+      val parkRef = db.collection(COLLECTION_PATH).document(pid)
+      val document = parkRef.get().await()
+
+      // Convert the document snapshot to a Park object
+      val park = document.toObject(Park::class.java)
+
+      if (park != null) {
+        // Check if the user has already rated
+        if (uid !in park.votersUIDs) {
+          // Calculate the new average rating and update the rating count
+          val updatedNbrRating = park.nbrRating + 1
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
```diff
@@ -142,7 +142,7 @@ class ParkRepositoryFirestore(private val db: FirebaseFirestore) : ParkRepositor
    */
   override suspend fun addRating(pid: String, rating: Int) {
```
Alex720p, line 143: Why are they 2 addRating fns and what's the purpose of adding a rating without being linked to a UID?

app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
```diff
@@ -294,6 +294,47 @@ class ParkRepositoryFirestore(private val db: FirebaseFirestore) : ParkRepositor
...
 
+  /**
+   * Add a rating to a park if the user has not already rated it.
+   *
+   * @param pid The park ID.
+   * @param uid The user ID of the person rating.
+   * @param rating The rating to add from 1 to 5.
+   */
+  override suspend fun addRating(pid: String, uid: String, rating: Float) {
+    require(pid.isNotEmpty()) { PID_EMPTY }
```
Alex720p, line 305: I might be missing something as the whole viewModel is structured like this and I haven't worked on it but:
What happens when the require fails? I haven't seen the exception being handled here or in the viewModel. Is it supposed to be handled in the UI? Might be worth adding it in the description of the function if that's the case

app/src/test/java/com/android/streetworkapp/model/park/ParkRepositoryFirestoreTest.kt
```diff
@@ -475,4 +475,59 @@ class ParkRepositoryFirestoreTest {
     assertEquals(5, park?.occupancy)
     assertEquals(listOf("event1", "event2"), park?.events)
   }
+
+  @Test
+  fun addRatingWithNewUserUpdatesRatingSuccessfully() = runTest {
```
Alex720p, line 480: Might be worth adding a test that tests the corner case when `val currentNbrRating = 1` to ensure that this case is handled if someone works on this code later on

Alex720p (changes requested):
Good code! 
I just have 2 questions that are below.
Side note: There is the code from Hotfix/park mvvm in this branch that I haven't reviewed. I guess you needed it to make the changes/forgot to remove it, might be worth merging the other branch first

tercierp:
> Good code! I have one improvement for the tests and have 2 questions that are below Side note: There is the code from Hotfix/park mvvm in this branch that I haven't reviewed. I guess you needed it to make the changes/forgot to remove it, might be worth merging the other branch first

Yep I'm waiting for the first branch to be reviewed than we will merge this code

app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
tercierp: This is a vestige of my basic UI debugging to see if the mvvm worked, good call I'm deleting it.

app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
tercierp: When `require` statement fails, it throws an `IllegalArgumentException` with the message `PID_EMPTY`, preconditions failing means that the developper is not using the function correctly, it's hard for me to see a case where it would be useful for the MVVM to "replace" a bad parameter with a good one without the developper being aware that its use is problematic.

app/src/test/java/com/android/streetworkapp/model/park/ParkRepositoryFirestoreTest.kt
tercierp: This test was added:
```kotlin
@Test
  fun deleteRatingWithCurrentNbrRatingOfOneResetsRatingToZero() = runTest {
    // Arrange
    val pid = "123"
    val rating = 4
    val currentRating = 4.0
    val currentNbrRating = 1

    // Mock Firestore interactions
    `when`(db.collection("parks")).thenReturn(collection)
    `when`(collection.document(pid)).thenReturn(documentRef)
    `when`(documentRef.get()).thenReturn(Tasks.forResult(document))
    `when`(document.getDouble("rating")).thenReturn(currentRating)
    `when`(document.getLong("nbrRating")).thenReturn(currentNbrRating.toLong())

    // Expected values after removing the rating
    val newNbrRating = 0
    val newRating = 0.0

    // Act
    parkRepository.deleteRating(pid, rating)

    // Assert
    verify(documentRef).update(mapOf("rating" to newRating, "nbrRating" to newNbrRating))
  }
```

tercierp:
> > Good code! I have one improvement for the tests and have 2 questions that are below Side note: There is the code from Hotfix/park mvvm in this branch that I haven't reviewed. I guess you needed it to make the changes/forgot to remove it, might be worth merging the other branch first
> 
> Yep I'm waiting for the first branch to be reviewed than we will merge this code

Since the previous PR is merged without any changes we can merge this PR when finished

app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
Alex720p: fair enough, wasn't sure if the app crashing was the expected behavior

Alex720p (approved):

PR #91, opened 2024-11-11
Feat/Use Nominatim API to display better park names

alvaro1080:
**Overview**

Until now, we displayed only a placeholder value for the name of each park. The objective of this Pull Request is to change those placeholder names with more meaningful names. To do so, we will display the road of each park by using the Nominatim API.

**Features**

- A new repository called `NominatimParkNameRepository` to utilise the Nominatim API
- An update to the UI that now uses this repository to display the road name of each park

**How to Test**

1. Run the Main application.
2. Click on any park.
3. The name of the park should now display the corresponding road name.

**Screenshot**

image


app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
```diff
@@ -18,6 +22,10 @@ open class ParkViewModel(private val repository: ParkRepository) : ViewModel() {
   val park: StateFlow<Park?>
     get() = _park
 
+  private val _parkLocation = MutableStateFlow<ParkLocation>(ParkLocation())
+  val parkLocation: StateFlow<ParkLocation>
+    get() = _parkLocation
+
```
SaturneV, line 28: I'm not sure I understand the purpose of this variable. The `parkLocation` is already stored in the `_currentPark`, isn't it?

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
```diff
@@ -102,6 +110,30 @@ open class ParkViewModel(private val repository: ParkRepository) : ViewModel() {
...
+  /**
+   * Set the new parkLocation
+   *
+   * @param parkLocation: a ParkLocation object
+   */
+  fun setParkLocation(parkLocation: ParkLocation?) {
+    require(parkLocation != null) { "ParkLocation can not be null" }
+    _parkLocation.value = parkLocation
+  }
+
```
SaturneV, line 136: Since this function is related to `_parkLocation`, I'd like to comment it to keep track. No need to reply.

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -62,6 +62,7 @@ fun MapScreen(
                 state = MarkerState(position = LatLng(park.lat, park.lon)),
                 onClick = {
                   parkViewModel.getOrCreateParkByLocation(park)
+                  parkViewModel.setParkLocation(park)
```
SaturneV, line 65: Here too, using the Park MVVM to set the current park already sets the ParkLocation. Wouldn't it be better to only use the current park?

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
```diff
@@ -70,6 +70,8 @@ fun ParkOverviewScreen(
 
   currentPark.value?.let { eventViewModel.getEvents(it) }
 
+  parkViewModel.updateCurrentParkNameNominatim()
```
SaturneV, line 73: I’m not sure of the usefulness of this function here. Every time a park is selected, it call the ViewModel function.

Wouldn’t it be simpler to call this function just once when a park is created, and then delegate this task to the ParkViewModel?

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
```diff
@@ -102,6 +110,30 @@ open class ParkViewModel(private val repository: ParkRepository) : ViewModel() {
...
+  fun updateCurrentParkNameNominatim() =
+      viewModelScope.launch {
+
+        // Only true when we have never updated the park name with nominatim to reduce calls to the
+        // API
+        if (_currentPark.value?.name?.contains("Default Park") == true) {
+          nameRepository.convertLocationIdToParkName(
+              _parkLocation.value.id,
+              { name -> _currentPark.value?.let { repository.updateName(it.pid, name) } },
+              {})
```
SaturneV, line 123: Logging in case of an error could be a good idea.

app/src/main/java/com/android/streetworkapp/model/park/ParkNameRepository.kt
```diff
@@ -0,0 +1,11 @@
+package com.android.streetworkapp.model.park
+
+/** A repository interface for park names. */
+interface ParkNameRepository {
```
SaturneV, line 4: Sonar raise an issue at this line: "[Make this interface functional or replace it with a function type.](https://sonarcloud.io/project/issues?pullRequest=91&open=AZMdlXonPZUF0SGa_CTm&id=SwEnt-Group8_Street-work-app)

SaturneV (changes requested):
Great PR that adds a useful feature! I’ll let you respond to my comments on the general implementation logic of your PR. Don’t forget to check the issues raised by Sonar as well. Also, consider adding more details in the description about the files and components you’ve modified. Good job!

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
alvaro1080: This increases the separation between the fetching of the name and the creation of the park, as describe bellow. (Also if you test your implementation, you will notice that it does not work. I have at the moment no explanation why. Have you manage to have an implementation that works using your idea?)

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
alvaro1080: I think it is better to separate the two, as creating the park is critical for our app, and it should rely as less as possible on other API. (Also as stated earlier, your implementation does not seem to work after a few tests)

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
alvaro1080: You are right I will update it, thank you!

app/src/main/java/com/android/streetworkapp/model/park/ParkNameRepository.kt
alvaro1080: I will fix it, thank you!

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
SaturneV: I understand your point of view, but I find it counterproductive to store the same variable twice. I've tried using the current park exclusively and managed to get a working version quickly and easily, we can talk about in person if you want!

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
SaturneV: I think in this case it's not a question of separation, given that this line itself doesn't call the API but just sets the location, as said above it just duplicates the location. From my side removing this line and only use the current park works correctly (see comment above).

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
SaturneV: I see why it was implemented like that then, I mark the comment as resolved, thank you!

app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
alvaro1080: I think it is a good idea, it is better to solve this problem in person.

SaturneV: After talking it over, we decided to leave the current implementation in place, as a change of implementation would require too much effort. I resolve this conversation and the ones related to.

SaturneV (approved):
Thank you for responding to my comments and for the improvements. After discussion we keep the current implementation and open a non-urgent issue #99 to refactor the Park MVVM in the future. I approve this PR it can be merged. Good job!

Issue #87, opened 2024-11-11
Bug: New User not being added in firebaseDb


Alex720p:
When a new user logs in he isn't being stored in the `users` collection on firebase:
Profile screenshot:
<img width="182" alt="Screenshot 2024-11-11 111652" src="https://github.com/user-attachments/assets/ba95a890-b013-460f-a00d-a85c077a73b8">
Firebase screenshot:
<img width="508" alt="Screenshot 2024-11-11 111855" src="https://github.com/user-attachments/assets/c18ff7a8-5171-4ab2-9361-3e1293cc412e">

Issue:
`AddUser` is never being called, it was called in `observeAndSetCurrentUser` but it is not used anymore
To recreate the issue, delete yourself from the database and login again. The entry won't be added.

PR #93, opened 2024-11-12
Feat/integrate progression backend into UI

Alex720p:
This pr aims to completely integrate the progression system in the app aswell as some final touches for the Progression UI.

- **`ui/Progression.kt`**
  - Integrate `ProgressionViewModel` and `UserViewModel` into UI
  - Retrieve progression values via the Progression ViewModel and display them into the UI
  - Add text if the users has no achievements to display
  - Assign more testTags

- **`model/Progression.kt`**
  - ***IMPORTANT LAST MINUTE FIX:*** added `getOrAddProgression` to the viewModel, the current implementation didn't work with our login flow. The `ProgressionViewModel` is not done in the state it is in but 'works', created https://github.com/SwEnt-Group8/Street-work-app/issues/112 to fix this next week. 
  - Modify `checkScore` function to make it more modulable if we add a different score system and/or different ranks
  - Integrate a soft ceiling for the maximum value `currentGoal` can take (via `checkScore` fn)

- **`screen/ProgressionTest.kt`**
  - Implement new tests with newly integrated viewmodels
 
- **`BottomNavigationMenu.kt`**
  - Set color value of Icons (only the material icons, so the 2 first starting from the right) to black, to match the figma design
 
- **`drawable/trophy_24px.xml`**
  - Comment out the `tint` param and set icon color value to black. On phones where darkmode was activated, the icon would turn white. As we have not implemented a dark/white mode system yet, the icon is hardcoded to black. Will need to be changed later

- **`SignIn.tk`**
  - Identify issue in signIn where new users weren't being added in the db. Created issue: https://github.com/SwEnt-Group8/Street-work-app/issues/87


app/src/main/java/com/android/streetworkapp/model/progression/Progression.kt
```diff
@@ -50,3 +50,9 @@ enum class MedalsAchievement(val achievement: Achievement) {
   GOLD(Achievement(R.drawable.place_holder, "Gold medal", emptyList(), "Third medal")),
   PLATINUM(Achievement(R.drawable.place_holder, "Platinum medal", emptyList(), "Last medal"))
```
tercierp, line 51: Suggestion: since figma already has the various images, it might be a good idea to put the right ones in instead of the placeholder. I don't know if it's the right PR for that but with figma you just have to right-click on the icon to copy it as a png.

app/src/main/java/com/android/streetworkapp/ui/navigation/BottomNavigationMenu.kt
```diff
@@ -66,7 +66,12 @@ fun BottomNavigationMenu(
                       )
                 }
                 topLevelDestination.icon?.let {
-                  Icon(topLevelDestination.icon, contentDescription = topLevelDestination.textId)
+                  Icon(
+                      topLevelDestination.icon,
+                      contentDescription = topLevelDestination.textId,
+                      tint =
+                          Color.Black) // same here as in the trophy xml, will need to modify this
```
tercierp, line 73: Note: good catch we need this modification for the presentation tomorrow

tercierp (changes requested):
One function's documentation is missing, the rest are comments and details.

I tested on my physical phone the mvvm works correctly even for a new user who has no progress at the moment !

Good work, thanks for this PR.

app/src/main/java/com/android/streetworkapp/model/progression/Progression.kt
Alex720p: I agree but I don't think those on figma really match the bronze, silver, gold, platinum system we have in place. We should probably get some new ones

tercierp (approved):
Good changes, thank you for your work

PR #102, opened 2024-11-13
Hotfix/handle query took too long from Nominatim

alvaro1080:
**Fix: Query took too long from Nominatim**
This Pull Request is used to handle the current response of Nominatim to every query, which is : "Query took too long to process." This response is a string that has not the JSON format, which makes our app crash. A try-catch is added to handle this situation

(We imagine that the API is currently down for every user, and that is why we have this response after we tried on different devices)

**How to test**
You can change the collection path of the parks to have a new fresh database, and check that the app do not crash on never seen before parks. (They also should have the default name, since Nominatim can not at the moment give the right name)


tercierp (approved):
Very clear, good patch, hoping that we haven't been banned from the Nominatim API and that this is only a temporary problem.

PR #103, opened 2024-11-13
Feat bluetooth for friends (backend part)

tercierp:
### Overview
This PR wasn't easy, so I'll try to explain and document my work as best I can to make it as easy to read as possible. **Do not hesitate to ask me a lot of questions**, it took me a while to understand how to use the bluetooth API. Many of the technical terms used in the code are taken from the [official bluetooth API documentation of developer.android.com](https://developer.android.com/develop/connectivity/bluetooth?hl=en). 

In order to reduce the size of the PR, I decided to make an initial PR focused solely on the functional aspect of the backend and the bare minimum of UI. The second PR will be UI-oriented

### Clarifications

The implementation has two main classes: `BluetoothServer` and `BluetoothClient`. The server corresponds to the phone that sends its `uid`, while the client is the second phone that receives the UI and makes a friend request via the `UserViewModel`.
Don't be confused by the terms server and client: both are phones. 
Simple diagram explaining how it works
<p align="center">
  image
</p>

### Technical choices
Implementing a [Gatt server and client](https://developer.android.com/reference/android/bluetooth/BluetoothGattServer#inherited-constants) allows small amounts of data to be sent using BLE (Bluetooth Low Energy) technology, thus limiting battery consumption. As soon as the user leaves the addFriend screen, the server and client are shut down to conserve the user's battery.

### Demo
Here's a test with a video capture of two physical phones

https://github.com/user-attachments/assets/74bcc202-2f1f-4182-88d2-c824d05771f6

The debug notification is not coded in the application, it's nRF connect which helped me a lot during the debugging to understand the different problems.


tercierp:
As I had thought it's very difficult to test the bluetooth API without modifying a lot of code, I've added a few easy-to-do tests but I can't really see how to test more.

app/src/main/java/com/android/streetworkapp/ui/profile/BluetoothButton.kt
```diff
@@ -0,0 +1,30 @@
...
+) {
+  val context = LocalContext.current
+
+  Button(
+      onClick = {
+        bluetoothController.startGattServer(uid) // Start GATT server to send UID
+        Toast.makeText(context, "Broadcasting UID via BLE GATT server", Toast.LENGTH_SHORT).show()
+      },
+      modifier = Modifier.size(220.dp, 50.dp).testTag("BluetoothButton")) {
+        Text(text = "Broadcast UID", fontSize = 17.sp)
```
alvaro1080, line 28: Suggestion: On successfully adding a friend, we could increase the score of each participant with the increase score function of the MVVM (note: this might be out of scope for this PR, you might want to do it in a future PR)

alvaro1080 (approved):
Thank you for this pull Request! Adding friends in a real life interaction was one of the main objective of our app, and it is great that you find the way to make it functional.

The bluetooth is a great alternative to the NFC, that created a lot of problem as we discussed during our meeting.

The code is clean, I did not find a lot to say about it. I also think that our line coverage is really good at the moment, so it is reasonable to validate this feature even with a less then 80% coverage, since it seem quite difficult to test. Adding friends is essential to our app , so it makes sens to reduce the coverage for such an essential feature.

app/src/main/java/com/android/streetworkapp/ui/profile/BluetoothButton.kt
tercierp: In second PR I will do the UI but I can also integrate the `score` update. Since I want to create a small pop-up dialog. So that the client can accept or refuse the friend request from the server. What do you think ?

Cf. [dialog documentation](https://developer.android.com/develop/ui/compose/components/dialog)

alvaro1080: I think it is the best way to do it, you are right!

PR #104, opened 2024-11-14
Feat/mvvm link to the UI for rating of parks

Simmanz:
### Implement the MVVM link for park rating

park-rating-before

park-rating-after

## Description of the changes

Adapted the code so that submitting a review will now correctly add it to the MVVM.

- Added `userViewModel` to the `ParkOverviewScreen()` **signature**.
- The `RatingButton` will now only display if the current user is not **null** (**connected**) and has **not rated the park yet**.
- Sanity checks in `handleRating()` to verfiy correct value / MVVM state beforehand.

## Tests

- Edited `ParkRatingTest` to also test that `RatingButton` is being correctly displayed.
- Created **[unitTest]** `parkRatingTest` - to test that `handleRating()` correctly calls (or not) the `addRating()` function.

## Note

Appearance might have changed slightly from images due to the merging with main (new PRs).


app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
```diff
@@ -253,6 +282,59 @@ fun RatingDialog(showDialog: MutableState<Boolean>) {
...
+    if (context != null)
+        Toast.makeText(context, "Invalid rating value, could not rate park", Toast.LENGTH_SHORT)
+            .show()
+  } else if (parkViewModel == null) {
+    if (context != null)
+        Toast.makeText(context, "Error with view model, could not rate park", Toast.LENGTH_SHORT)
+            .show()
+  } else {
+    Log.d("ParkOverview", "handleRating: Adding rating to park")
+    if (context != null) Toast.makeText(context, "Rating submitted", Toast.LENGTH_SHORT).show()
```
kzepfl, line 319: Instead of calling context != null everywhere call it once at the start

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
```diff
@@ -253,6 +282,59 @@ fun RatingDialog(showDialog: MutableState<Boolean>) {
...
+fun handleRating(
+    context: Context?,
+    park: Park?,
+    user: User?,
+    starRating: Int,
+    parkViewModel: ParkViewModel?
+) {
+  Log.d("ParkOverview", "handleRating: {park=$park ; user=$user ; rating=$starRating")
+
+  if (user == null) {
```
kzepfl, line 303: Maybre you can use kotlin when function instead of if else, i find it more readable.
https://www.educba.com/kotlin-when-multiple-conditions/

app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/ParkRatingTest.kt
```diff
@@ -108,4 +112,60 @@ class ParkRatingTest {
...
+    composeTestRule.onNodeWithTag("ratingButton").assertIsDisplayed()
+  }
+
+  @Test
+  fun isRatingButtonCorrectlyDisplayedWhenRating() {
+    val showRatingDialog = mutableStateOf(false)
+    val user = User("uid", "username", "email", 0, emptyList())
+
+    // default park constructor
+    val emptyPark =
```
kzepfl, line 141: you used showRatingDialog and emptypark twice, so initiating it in a @before `setup function` would be a good practice

app/src/test/java/com/android/streetworkapp/model/park/ParkRatingTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  @Test
+  fun handlerCorrectlyCallingAddRatingForParkValues() = runTest {
+    val starRating = 1
+    val user = User("uid", "username", "email", 0, emptyList())
+
+    // Testing for park with no rating and park with rating :
+    val emptyPark =
```
kzepfl, line 91: Here aswell all the empty park are the same, it can be init in a @before` setup function`

kzepfl (commented):
Good PR, the comment is very helpfull. There is just the if-else statement that could be changed and some value in test that should be in a @before setup function

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
Simmanz: ### Why the null checks ?
The reason for the null checking is that each `Toast()` instance require a `context`. Because I want to test this component (`handleRating`) and it is not a **@composable** function, it has no associated `context` by itself (cannot fetch it using `localContext.current` for instance). Meaning I need to **pass it from the UI which calls it.** 

At the same time when I test the component, I cannot give it a non-null `context` as it is a **unit test** (not related to UI) and I therefore cannot create the `context` from such a test that does not call any @composable function.

### The decision I took
In order to keep the Toast functionality while being able to test the function, I choose to accept the passing of a null-context and to simply not display them when it is null (i.e when we are running the unit test).

### Changes

If I were to add a context check at the start it would imply that any rating submission would be denied if the context is null. I do not know if there are normal instances in which it could. Therefore I was reticent about doing so.

I can propose to add a **parameter** to the function to make the function silent, which would replace `context` checks by `silent` boolean checks if you want ?

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
Simmanz: Whens would not make it more redable in this case because the variable is never the same : 

`when(user){
  null -> {Toast}
  else -> {
    when(park){
      null -> {Toast}
      else -> {
        .....
      }
    }
}
}`

Would it be better for your readability to remove the else of the else if (and add `return@handleRating` inside each ifs) would make it better by allowing more space between each check by providing less nesting though ?

kzepfl: sorry it was not a good link 
https://kotlinlang.org/docs/control-flow.html#when-expressions-and-statements
I meant somethin like this 

```
when {
    a > b -> "a is greater than b"
    a < b -> "a is less than b"
    else -> "a is equal to b"
}
```
it check if statement is true

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
kzepfl: I don't understand why you can't check the context before the if-else statement.
You check the context in every single if-else statement, if the context is null to begin with you don't even need to check all the if-else statement.

Something like this
```
if (context != null){
  if (user == null) {
          Toast.makeText(context, "User not found, could not rate park", Toast.LENGTH_SHORT).show()
    } else if (park == null) {
          Toast.makeText(context, "Park not found, could not rate park", Toast.LENGTH_SHORT).show()
    }
}
```

> The reason for the null checking is that each Toast() instance require a context

or do you mean that each Toast have a different context and if it is then yes you would need to check context seperatly

app/src/androidTest/java/com/android/streetworkapp/ui/parkoverview/ParkRatingTest.kt
kzepfl: Empty park can be put inside the setup function

kzepfl (commented):
There are still some changes in test that is not made and
my bad, I linked a bad site to demanstrated `when,` I added a new one.

I dont quite understand the context probleme, so i explicited my reasonning. If my reasoning is false then we can keep the context check and its approved :D

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
Simmanz: Ok ! I understand what you meant by when now ! I changed it :)

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
kzepfl: I sawed that in the else statement you had parkViewModel.addRating(park.pid, user.uid, starRating.toFloat()) that neede to be called at all time, so i understand why ou can't put context check at the start

Simmanz: Each Toast need the same context. What I meant by not being able to check before the if-else statement is that it would modify the functionality of the handling of rating of parks (no longer only about Toasts). 

The only other possibility I see is to make the function return Int error codes which would be handled in the UI to display the Toasts, there would be no passing of context anymore but the complexity would increase a lot.

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
kzepfl: It was purely for eshtetic if you find if-else more appropriate it can be changed

kzepfl (approved):
I undestood the Context probleme and all my issue were addressed.

PR #110, opened 2024-11-14
Feat/display user on the map

kzepfl:
### Display the user on the map using google API

image

## Changes

- Changed `Map.kt`, now it will ask for permission when the app is booted and if permssion is acquired it will update the location to where the user is.
- Changed `MapUi.kt` to display the user as a blue dot on the map
- Added 2 new classes in utils : `LocationService.kt` and `PermissionManager` abstraction of function used in `Map.kt`
_`LocationService.kt` : if permission is granted, it update the user location_
_`PermissionManage` : manage the permission, can verify permission and ask for it_

    Changed the image in Profile.kt to display the current user profile picture
    Added a dependecies in build.gradle.kts for coil (needed to display the profile picture)

## Testing

- Granted location permission inside `MapUiTest.kt`, it would crashes otherwise
- Added test for mapManager function of `Map.kt` (the coverage is only at 78%)


Alex720p (approved):
Looks good.
Tests could be in more depth but I'm guessing it's hard to test/mock extensively these types of things

PR #115, opened 2024-11-14
Test/end 2 end to create event

misterM125:
**Overview**
This PR adds a new end to end test that simulates a user that creates an event

The steps are as follows:
- the user logs in the app via google authentication (the authentication part is skipped as i will assume google authentication works)
- the user then clicks on a marker to access a park
- he will then click on the create event button in the park overview to create a new event
- he will then be redirected to the park overview screen where he will be able to see his new event displayed
- he will finally click on the "about" button of the event to see its details displayed on the event overview screen

**Additional info**
A few composables have been modified to enable easier testing:
- the `parkrepository` now has a `testing` parameter to change the collection used to store parks on firestore ( this ensures that the the public park domain will not uselessly be modified and allows for more consistent testing)
- A `LaunchedEffect` has been added to the `mainActivity` wrapping the `testInvokation` to easily change between screens during testing
- gradle file has been changed to exclude protobuf (Testing the mvvm doesn't work with it don't ask me why)
- I also added a small test to `mapUITest` as i figured out how to test marker behavior


app/src/main/java/com/android/streetworkapp/MainActivity.kt
```diff
@@ -222,7 +221,9 @@ fun StreetWorkApp(
               }
             }
 
-        if (firstTimeLoaded || navTestInvokationOnEachRecompose) {
+        if (e2eEventTesting) {
```
Alex720p, line 224: I did somethng similar in my class, we'll have to choose one or the other

Alex720p (approved):
Looks good.
The timings check in the tests are not ideal but I guess there's no other way around for testing with google map

PR #126, opened 2024-11-18
Refactor/Update and finish the rework of the progression MVVM 

alvaro1080:
**Overview**
Since the Login system changed with a new way to set the current user, a new function was added to the Progression ViewModel : `getOrAddProgression` . Since this function is a merge of the `getProgression` and the `createProgression`, this Pull Request will remove them since they are obsolete. This Pull Request also include a first instance of the gamification system, in order to check that the progression works as intended after this refactoring.

**Changes**
- Removed `getProgression` the `createProgression`
- Use `increaseUserScore` in `AddEvent`, to have our first instance of gamification, and to use/test the progression Screen

**How to Test**

1. Log in the app
2. Select any park
3. Add an event to this Park, you should receive points
4. Go to the progression screen and check that the score is higher now

(Note: the number of points given is there to easily see a difference in the progression screen, a dedicated new PR should be dedicated to use `increaseUserScore` in every places needed and with a coherent number of points)

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
+++ b/app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
@@ -52,6 +52,7 @@ import com.android.streetworkapp.model.event.Event
 import com.android.streetworkapp.model.event.EventConstants
 import com.android.streetworkapp.model.event.EventViewModel
 import com.android.streetworkapp.model.park.ParkViewModel
+import com.android.streetworkapp.model.progression.SCORE_INCREASE
 import com.android.streetworkapp.model.user.UserViewModel
 import com.android.streetworkapp.ui.navigation.NavigationActions
 import com.android.streetworkapp.ui.theme.ColorPalette
@@ -131,7 +132,7 @@ fun AddEventScreen(
               parkViewModel.addEventToPark(event.parkId, event.eid)
 
               // Used for the gamification feature
-              userViewModel.increaseUserScore(event.owner, 510)
+              userViewModel.increaseUserScore(event.owner, SCORE_INCREASE.CREATE_EVENT.scoreAdded)
               // Note: temporary value to use the progression screen. Should be update once
               // the gamification is completed
               Toast.makeText(context, "+510 Points", Toast.LENGTH_SHORT).show()
--- a/app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
+++ b/app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
@@ -135,7 +135,7 @@ fun AddEventScreen(
               userViewModel.increaseUserScore(event.owner, SCORE_INCREASE.CREATE_EVENT.scoreAdded)
               // Note: temporary value to use the progression screen. Should be update once
               // the gamification is completed
-              Toast.makeText(context, "+510 Points", Toast.LENGTH_SHORT).show()
+              Toast.makeText(context, "+"+SCORE_INCREASE.CREATE_EVENT.scoreAdded+" Points", Toast.LENGTH_SHORT).show()
 
               navigationActions.goBack()
             }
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
```diff
@@ -75,6 +75,7 @@ fun ProgressScreen(
   val currentProgression by progressionViewModel.currentProgression.collectAsState()
 
   currentUser?.uid?.let { progressionViewModel.getOrAddProgression(it) }
+  currentUser?.let { progressionViewModel.checkScore(it.score) }
```
Alex720p, line 78: There's an issue here. Imagine a user collects enough point to pass 2 levels, this would only grant him the first achievement and not the second one. If we still want to keep the user viewmodel and progression viewmodel we should have a loop in the checkscore that loops until the user has been granted all achievements

app/src/test/java/com/android/streetworkapp/model/progression/ProgressionRepositoryFirestoreTest.kt
```diff
@@ -70,52 +69,6 @@ class ProgressionRepositoryFirestoreTest {
     assertEquals("uniqueProgressionId", progressionId)
   }
```
Alex720p, line 71: Suggestion: There is no test on getOrAddProgression that checks behavior when the document exists or not. Not super important, if there's other pressing things to do we can leave it as is

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -130,6 +130,12 @@ fun AddEventScreen(
               eventViewModel.addEvent(event)
               parkViewModel.addEventToPark(event.parkId, event.eid)
 
+              // Used for the gamification feature
+              userViewModel.increaseUserScore(event.owner, 510)
```
Alex720p, line 134: Improvement: instead of having hardcoded values for the score increase we could have an enum SCORE_INCREASE(val increase:Int), this would make things much more maintanable

Alex720p (changes requested):
There's a few things that needs addressing:

 - The scores increases should probably be in an enum to increase code maintanability
 - The checkScore function should be looped over until user has received all his achievements. (or called on each score increase)
 - getOrAddProgression's behavior is not tested for when a document exists or not. (This one is really not important, you can leave it as is)

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
alvaro1080: Good catch, I added the loop in `checkScore` (in the ViewModel, since it makes more sens than in the UI. Thank you!

app/src/test/java/com/android/streetworkapp/model/progression/ProgressionRepositoryFirestoreTest.kt
alvaro1080: This PR is in fact a prelude for the actual task of this week (which is adding new features to the progression screen), therefore I think it will make more sens to make this update in the upcoming PR with all the other changes that will be made along the way.

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
alvaro1080: Done, thank you!

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
Alex720p: There's a slight modification I would make to the loop: instead of hardcoding platinum as a condition, you could add a ACHIEVEMENT_CEILING (or something like that) to the enum. Then you could just fetch the max achievement by taking ordinal ACHIEVEMENT_CEILING - 1. This way if we add ranks, everything still works

alvaro1080: I am not 100% sure that I see how you imagine it, but I updated the progression with a CEILING value. You can check it and tell me if you disagree.

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
alvaro1080: It is fix, thank you!

Alex720p (approved):

PR #128, opened 2024-11-18
Feat/implement text moderation with perspective api

Alex720p:
This pr aims to implement a text moderation system for our app.

The current flow works like this:
I have a text to be analyzed called `content`, I call `TextModerationViewModel::analyzeText` with `content`. I can select 2 functions as callbacks: one if the the text is under the thresholds (considered clean), another one if the text is over the thresholds. I can either provide the threshold values by myself or use the default ones.

**Note:** I haven't tinkered too much around the default threshold values. . I put the values quite low at 0.2 (Probability of text being of tag 'x') since the only use cases are for event names and event descriptions (and those texts should be non-ambiguous).
The system is far from perfect. If I have some long inputs with a few words of hate speech only at the end, this won't get picked up.

- **`libs.version.toml.kt` & `build.gradle.kts`**
  - Add BuildConfig to be able to retrieve the Perspective API key from local.properties
  - Add dependency for kotlinxSerializationJson that let's us easily serialize/deserialize api requests/responses

- **`TextModerationRepository.kt`**
  - Interface for our TextModeration repository

- **`PerspectiveAPIModels.kt`**
  - Contains all the data structures needed to interact with the API as well as some custom types for return values and tags

- **`PerspectiveAPIRepository.kt`**
  - Wrapper for the Perspective API. Handles all requests to the api

- **`TextModerationViewModel.kt`**
  - ViewModel for our TextModeration repository

- **`PerspectiveAPIRepositoryTest.kt` & `TextModerationViewModelTest.tk`**
  - Tests for the repository and viewmodel respectively

**Note:** It is not implemented in the UI, this is for another pr

**Update:** There is a rate limit of 1 RPS for the API. If we plan to release app, i'll have to take this into account, for now this is unhandled.

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
@@ -29,6 +29,14 @@ data class Comment(
 
 //Responses data classes
 @Serializable
+data class SuccessResponse(
+    val attributeScores: Map<String, AttributeScore>,
+    val languages: List<String>,
+    val detectedLanguages: List<String>
+)
+
+
+@Serializable
 data class AttributeScore(
     val spanScores: List<SpanScore>,
     val summaryScore: Score
@@ -116,5 +124,13 @@ enum class PerspectiveApiErrors(val errorMessage: String) {
      */
     JSON_DESERIALIZATION_ERROR("Failed to deserialize input from Perspective API response"),
 
-    UNSUPPORTED_HTTP_CODE("Received an unsupported HTTP code")
+    /**
+     * Received a HTTP code not included in {200, 400}
+     */
+    UNSUPPORTED_HTTP_CODE("Received an unsupported HTTP code"),
+
+    /**
+     * Received an empty body as API response
+     */
+    EMPTY_BODY_RESPONSE("Received an empty body in Perspective API response")
 }
--- a/app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
@@ -2,38 +2,31 @@ package com.android.streetworkapp.model.moderation
 
 import kotlinx.serialization.Serializable
 
+// All the definitions below are used to interact with Perspective API
 
-//All the definitions below are used to interact with Perspective API
-
-/**
- * Default values we can use as thresholds for our TextModerationTags
- */
-object PerspectiveAPIThresholds{
-    val DEFAULT_THRESHOLD_VALUES = TextModerationTags.entries.associateWith { 0.5 }
+/** Default values we can use as thresholds for our TextModerationTags */
+object PerspectiveAPIThresholds {
+  val DEFAULT_THRESHOLD_VALUES = TextModerationTags.entries.associateWith { 0.5 }
 }
 
-/**
- * All the tags we want to fetch from the API
- */
+/** All the tags we want to fetch from the API */
 enum class TextModerationTags {
-    TOXICITY, INSULT, THREAT
+  TOXICITY,
+  INSULT,
+  THREAT
 }
 
-/**
- * Used to store the results from the API
- */
+/** Used to store the results from the API */
 data class TagAnnotation(val tag: TextModerationTags, val probability: Double)
 
-
-/**
- * Used to return the result of a query to PerspectiveAPI
- */
+/** Used to return the result of a query to PerspectiveAPI */
 sealed class TextEvaluationResult {
-    data class Success(val annotations: List<TagAnnotation>) : TextEvaluationResult()
-    data class Error(val errorType: PerspectiveApiErrors) : TextEvaluationResult()
+  data class Success(val annotations: List<TagAnnotation>) : TextEvaluationResult()
+
+  data class Error(val errorType: PerspectiveApiErrors) : TextEvaluationResult()
 }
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
```diff
@@ -0,0 +1,107 @@
...
+  val DEFAULT_THRESHOLD_VALUES = TextModerationTags.entries.associateWith { 0.2 }
+}
+
+/** All the tags we want to fetch from the API */
+enum class TextModerationTags {
+  TOXICITY,
+  INSULT,
+  THREAT,
+  PROFANITY
+}
```
alvaro1080, line 18: Question: Is there a way to identify https links? I think avoiding advertising/harmful links could also be useful.

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+  private fun getTextAnnotations(content: String): TextEvaluationResult {
+    // I don't perform any checks on the content, I'll let the api handle the error checking
+
+    val requestMediaType = "application/json; charset=utf-8".toMediaType()
+    val requestBody = this.formatPostRequestBody(content).toRequestBody(requestMediaType)
+
+    // note: someone could get the key by sniffing the packets, but since we can't have a backend
+    // here we are :)
+    val request =
+        Request.Builder()
```
alvaro1080, line 57: Suggestion: There is a "cleaner" way to build a URL. Here is for instance what we used for `Nominatim` : 

```
val url =
        HttpUrl.Builder()
            .scheme("https")
            .host("nominatim.openstreetmap.org")
            .addPathSegment("lookup")
            .addQueryParameter("osm_ids", "W$locationId")
            .addQueryParameter("format", "json")
            .build()
```

(Note: this is non critical, if it creates any problems with this specific query you can ignore this suggestion)

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
```diff
@@ -0,0 +1,107 @@
...
+
+// All the definitions below are used to interact with Perspective API
+
+/** Default values we can use as thresholds for our TextModerationTags */
+object PerspectiveAPIThresholds {
+  val DEFAULT_THRESHOLD_VALUES = TextModerationTags.entries.associateWith { 0.2 }
+}
+
+/** All the tags we want to fetch from the API */
+enum class TextModerationTags {
```
alvaro1080, line 13: Question: Is there a way to use this API also for images? This might be useful in the future for some features (in particular, adding a picture to a given park)

alvaro1080 (changes requested):
Thank you for this Pull Request! It is essential that we check the content of the inputs of the user to have a safe App.
The code is clean, and it is a great feature that is more than necessary in terms of moderation. 
I have a few questions and comments that I would like to discuss:

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
Alex720p: Nope, but I'm pretty sure I could implement something myself that does that. I'll see what I can do.

Here's all the tags we can use from the API if you're interrested: https://developers.perspectiveapi.com/s/about-the-api-attributes-and-languages?language=en_US
There's a thing for spam that could maybe work for links? But it's only for english

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
Alex720p: No, there's google api vision that we could use though. It would be a great addition, I could start working on it at the end of the week.

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIRepository.kt
Alex720p: Didn't know about it, it's done

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIModels.kt
Alex720p: I've looked around, I could probably do something with Regex but there's several drawbacks: Either I miss some links, or I get them all and there's a bunch of false positives. As links aren't clickable by default in Texts in Android (from what I've seen), I don't think it's worth the time. We could put a little disclaimer the first time a user opens an event/logs in though

alvaro1080: I would argue that "get all links" is perhaps the best solution, as there is probably, in our app, no good reason to let a user post any link. But I think we are going outside the scope of this PR.

Alex720p: Yeah that's something for another pr. I think it's complicated to get all links without ruining a part of the user experience. 
If you look at this page for example: https://fr.wikipedia.org/wiki/Liste_des_domaines_Internet_de_premier_niveau
I can easily see happening that a user types '${word}.${some domain name}' by mistake because he forgot to add a space after ponctuation.

alvaro1080 (approved):
Thank you for the updates and the answers!

PR #133, opened 2024-11-19
Feat/join event

misterM125:
**Overview**
This PR implements the ability to join / leave an event from the overview screen, which will then update the database and as such, the event's displayed information on the UI

**Implementation**
- The functions `addParticipantToEvent` and `removeParticipantFromEvent` have been added to the event mvvm
- The event overview `bottomBar` has been modified to reflect these changes:
- the button of the `bottomBar` will now call the mvvm when clicked on to perform certain actions and changes depending if the user is already part of the event or not
- A button "join this event" will be showed to a user who is not part of the event yet and a button "leave this event" will be showed to a user who has already joined the event
- The owner of the event will automatically be added to the list of participants
- The owner of an event should not be able to quit his own event so the button will be disabled for him

**Testing**
These functionalities can be tested by browsing and creating events from multiple accounts so you can see how the screen looks if you're part of an event or not

Screenshot from 2024-11-19 19-31-46

Screenshot from 2024-11-19 19-32-01

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
@@ -68,6 +68,62 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
     }
   }
 
+  /**
+   * Add a participant to an event.
+   *
+   * @param eid The event ID.
+   * @param uid The user ID.
+   */
+  override suspend fun addParticipantToEvent(eid: String, uid: String) {
+    require(eid.isNotEmpty())
+    require(uid.isNotEmpty())
+    try {
+      val event = getEventByEid(eid)
+      if (event != null) {
+        val updatedParticipants = event.participants + 1
+        val updatedListParticipants = event.listParticipants.toMutableList()
+        updatedListParticipants.add(uid)
+        db.collection(COLLECTION_PATH)
+            .document(eid)
+            .update(
+                "participants", updatedParticipants, "listParticipants", updatedListParticipants)
+            .await()
+      } else {
+        Log.e("FirestoreError", "Error adding participant to event: Event does not exist.")
+      }
+    } catch (e: Exception) {
+      Log.e("FirestoreError", "Error adding participant to event: ${e.message}")
+    }
+  }
+
+  /**
+   * Remove a participant from an event.
+   *
+   * @param eid The event ID.
+   * @param uid The user ID.
+   */
+  override suspend fun removeParticipantFromEvent(eid: String, uid: String) {
+    require(eid.isNotEmpty())
+    require(uid.isNotEmpty())
+    try {
+      val event = getEventByEid(eid)
+      if (event != null) {
+        val updatedParticipants = event.participants - 1
+        val updatedListParticipants = event.listParticipants.toMutableList()
+        updatedListParticipants.remove(uid)
+        db.collection(COLLECTION_PATH)
+            .document(eid)
+            .update(
+                "participants", updatedParticipants, "listParticipants", updatedListParticipants)
+            .await()
+      } else {
+        Log.e("FirestoreError", "Error removing participant from event: Event does not exist.")
+      }
+    } catch (e: Exception) {
+      Log.e("FirestoreError", "Error removing participant from event: ${e.message}")
+    }
+  }
+
   fun documentToEvent(document: DocumentSnapshot): Event? {
     return if (document.exists()) {
       Event(
--- a/app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
@@ -4,6 +4,7 @@ import android.util.Log
 import com.android.streetworkapp.model.park.Park
 import com.google.firebase.Timestamp
 import com.google.firebase.firestore.DocumentSnapshot
+import com.google.firebase.firestore.FieldValue
 import com.google.firebase.firestore.FirebaseFirestore
 import kotlinx.coroutines.tasks.await
 
@@ -12,6 +13,8 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
 
   companion object {
     private const val COLLECTION_PATH = "events"
+    private const val FIELD_PARTICIPANTS = "participants"
+    private const val FIELD_LIST_PARTICIPANTS = "listParticipants"
   }
 
   /**
@@ -80,13 +83,13 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
     try {
       val event = getEventByEid(eid)
       if (event != null) {
-        val updatedParticipants = event.participants + 1
-        val updatedListParticipants = event.listParticipants.toMutableList()
-        updatedListParticipants.add(uid)
         db.collection(COLLECTION_PATH)
             .document(eid)
             .update(
-                "participants", updatedParticipants, "listParticipants", updatedListParticipants)
+                FIELD_PARTICIPANTS,
+                FieldValue.increment(1),
+                FIELD_LIST_PARTICIPANTS,
+                FieldValue.arrayUnion(uid))
             .await()
       } else {
         Log.e("FirestoreError", "Error adding participant to event: Event does not exist.")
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -68,6 +68,62 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+    try {
+      val event = getEventByEid(eid)
+      if (event != null) {
+        val updatedParticipants = event.participants + 1
+        val updatedListParticipants = event.listParticipants.toMutableList()
+        updatedListParticipants.add(uid)
+        db.collection(COLLECTION_PATH)
+            .document(eid)
+            .update(
+                "participants", updatedParticipants, "listParticipants", updatedListParticipants)
```
tercierp, line 89: Improvement: `updatedParticipants` can lead to race conditions (if multiple clients try to update the event at the same time) instead I discover the `FieldValue.increment(1)`. With this change you can remove the line 83.

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -68,6 +68,62 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+   * @param eid The event ID.
+   * @param uid The user ID.
+   */
+  override suspend fun addParticipantToEvent(eid: String, uid: String) {
+    require(eid.isNotEmpty())
+    require(uid.isNotEmpty())
+    try {
+      val event = getEventByEid(eid)
+      if (event != null) {
+        val updatedParticipants = event.participants + 1
```
tercierp, line 83: Improvement: As said below you can remove this line

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -68,6 +68,62 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+    try {
+      val event = getEventByEid(eid)
+      if (event != null) {
+        val updatedParticipants = event.participants - 1
+        val updatedListParticipants = event.listParticipants.toMutableList()
+        updatedListParticipants.remove(uid)
+        db.collection(COLLECTION_PATH)
+            .document(eid)
+            .update(
+                "participants", updatedParticipants, "listParticipants", updatedListParticipants)
```
tercierp, line 117: Suggestion: As done in `ProgressionRepositoryFirestore` you use constant for "participants" and "listParticipants"
Like this
```kotlin
companion object {
    const val FIELD_PARTICIPANTS = "participants"
    const val FIELD_LIST_PARTICIPANTS = "listParticipants"
}
```

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -68,6 +68,62 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+   * @param eid The event ID.
+   * @param uid The user ID.
+   */
+  override suspend fun removeParticipantFromEvent(eid: String, uid: String) {
+    require(eid.isNotEmpty())
+    require(uid.isNotEmpty())
+    try {
+      val event = getEventByEid(eid)
+      if (event != null) {
+        val updatedParticipants = event.participants - 1
```
tercierp, line 111: Improvement: As said below you can remove this line

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
tercierp, line 117: Improvement: `updatedParticipants` can lead to race conditions (if multiple clients try to update the event at the same time) instead I discover the `FieldValue.increment(-1)`. With this change you can remove the line 111.

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -68,6 +68,62 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+        val updatedListParticipants = event.listParticipants.toMutableList()
```
tercierp, line 112: Improvement: I think you could use `FieldValue.arrayRemove` instead of creating unnecessary copie of the list

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -68,6 +68,62 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+   * @param uid The user ID.
+   */
+  override suspend fun addParticipantToEvent(eid: String, uid: String) {
+    require(eid.isNotEmpty())
+    require(uid.isNotEmpty())
+    try {
+      val event = getEventByEid(eid)
+      if (event != null) {
+        val updatedParticipants = event.participants + 1
+        val updatedListParticipants = event.listParticipants.toMutableList()
```
tercierp, line 84: Improvement: I think you could use `FieldValue.arrayUnion` instead of creating unnecessary copie of the list

app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
```diff
@@ -230,6 +236,50 @@ fun DashBoardBar() {
...
+
+  Button(
+      onClick = {
+        Toast.makeText(context, "You have left this event", Toast.LENGTH_LONG).show()
+        eventViewModel.removeParticipantFromEvent(event.eid, user.uid)
+        navigationActions.goBack()
+      },
+      enabled = event.owner != user.uid,
+      modifier = Modifier.testTag("leaveEventButton"),
+      colors = ColorPalette.BUTTON_COLOR.copy(containerColor = Color.Red)) {
```
tercierp, line 278: Suggestion: I don't think the color of this button is appropriate. It seems to me that bright red is used to display errors, so if I were you I'd use the classic button color. And if you want to draw the user's attention, you can create an alert when he clicks with `Dialog`. What do you think ?

tercierp (changes requested):
Thanks for your good work! I've found a few improvements to optimize user experience and application performance. If you have any questions, don't hesitate to contact me.

app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
```diff
@@ -230,6 +236,50 @@ fun DashBoardBar() {
...
+    event: Event,
+    eventViewModel: EventViewModel,
+    user: User,
+    navigationActions: NavigationActions
+) {
+  val context = LocalContext.current
+
+  Button(
+      onClick = {
+        Toast.makeText(context, "You have joined this event", Toast.LENGTH_LONG).show()
```
alvaro1080, line 250: Suggestion: you could use `IncrementScore` from the user MVVM in order to give points when a user joints an event, since the progression MVVM now can display the increment of score. This would be great for the gamification part of the app!

(Note: you can have inspiration from the addEvent screen in the current main branch, that uses `increaseScore` and displays a toast with the number of points obtained)

alvaro1080 (changes requested):
In addition to the previous comments, I would like to add a suggestion:

app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
misterM125: The goal of making the button to leave an event ("cancel" action so to speak) red was essentially to differentiate it from the button to join an event ("enter" action)
I think making the buttons two different colors improves readability and helps alert users that the button's action has changed even before he clicks on it. The "cancel" button of the rating popup on the park overview screen uses red for example so i don't think we need to strictly follow a rule of red being used only for errors (as long as we don't go overboard)
That is why i think making both buttons a different color is necessary, let me know what you think!

app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
misterM125: I think that's a good idea but might be exploited if not made carefully:
what i mean by that is that a user can easily leave an event and join it back again to increase his score continuously.
A user could also join as many events as possible and never show up to them, there are probably other ways this could be exploited that i'm not thinking about right now.
That is why I think we should discuss together the overall gamification of the app and how we want to implement it and make it safe before simply adding some `IncrementScore` functions haphazardly.
As such this might be a job for a potential gamification centered branch instead of this one

alvaro1080: I understand your concern. It was more about having something to show on the next demo, but I see your point. Since there is a lot of other part of the app that is not yet ready for such adversarial behavior (and it is expected as it is not a finished product), I think focusing on the next demo could make sense, but it is up to you.

alvaro1080 (approved):
I approve the PR, since my suggestion has been answered in comment.

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
misterM125: good call! this has been changed

tercierp (approved):

tercierp:
Good changes thank you

PR #135, opened 2024-11-19
Feat/workout-MVVM

tercierp:
## Overview
This PR introduces the structure required to track and store all the training information of a User. Tracking user training data is critical for improving the app's user experience and enabling advanced analytics. The structure is designed to accommodate different types of training sessions while ensuring scalability.

## Data structure
The data structure is pretty complex but there is multiple reason why:

- User-Centric MVVM: Each `WorkoutData` is related to a uid. With this structure, you can retrieve all workout data using the uid of the user.

A natural question might be: why not update the existing UserViewModel? This decision is based on performance concerns. In our current architecture, the UserViewModel is accessed in many parts of the project. Adding a significant amount of new data per user would unnecessarily increase memory usage, as much of this data would go unused in other parts of the project.

- Flexible Data Model: Designing the data structure was non-trivial, as we needed to create a model applicable to multiple scenarios (solo trainings, referee trainings, and challenges).

1. The top level `WorkoutData` should contains all the workout sessions (`WorkoutSession`) done by an `User`
```kotlin
data class WorkoutData(
    val userUid: String = "",
    val workoutSessions: List<WorkoutSession> = emptyList()
)
```
2. The mid level `WorkoutSession` should contains all the data related to a session with the list of `Exercise`
```kotlin
data class WorkoutSession(
    val sessionId: String,
    val startTime: Long,
    val endTime: Long,
    val sessionType: SessionType,
    val participants: List<String> = emptyList(),
    val exercises: List<Exercise> = emptyList(),
    val winner: String? = null
)

/** An enum class representing the type of workout session. */
enum class SessionType {
  ALONE, // Solo session
  REFEREE, // Referee-observed session
  CHALLENGE // Competitive session
}
```
3. The bottom level the `Exercise` contains all the data concerning one specific execice done by the user with a specific time 
```kotlin
data class Exercise(
    val name: String,
    val reps: Int? = null,
    val sets: Int? = null,
    val weight: Float? = null,
    val duration: Int? = null
)
```

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/model/workout/WorkoutRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/workout/WorkoutRepositoryFirestore.kt
@@ -64,28 +64,38 @@ class WorkoutRepositoryFirestore(private val db: FirebaseFirestore) : WorkoutRep
    *
    * @param uid The UID of the user (also the WorkoutData ID).
    * @param sessionId The ID of the session to update.
-   * @param exercises A list of exercises to add or update.
-   * @param endTime The end time of the session.
+   * @param exercises A list of exercises to add or update, if empty the exercises are not updated.
+   * @param endTime The end time of the session, if null the end time is not updated.
+   * @param winner The winner of the session, if null the winner is not updated.
+   *
    */
   override suspend fun updateWorkoutSessionDetails(
       uid: String,
       sessionId: String,
       exercises: List<Exercise>,
-      endTime: Long?
+      endTime: Long?,
+      winner: String?
   ) {
     require(uid.isNotEmpty()) { ERROR_UID_EMPTY }
     require(sessionId.isNotEmpty()) { ERROR_SESSION_ID_EMPTY }
 
     try {
-      // Construct the field path for the specific session's exercises and endTime
-      val sessionFieldPath = WORKOUT_SESSIONS
-      val updateMap =
-          mapOf(
-              "$sessionFieldPath.$sessionId.exercises" to exercises,
-              "$sessionFieldPath.$sessionId.endTime" to endTime)
+      // Construct the field path for the specific session details
+      val updateMap = mutableMapOf<String, Any?>()
 
+      if (exercises.isNotEmpty()) {
+        updateMap["$WORKOUT_SESSIONS.$sessionId.exercises"] = exercises
+      }
+      if (endTime != null) {
+        updateMap["$WORKOUT_SESSIONS.$sessionId.endTime"] = endTime
+      }
+      if (!winner.isNullOrEmpty()) {
+        updateMap["$WORKOUT_SESSIONS.$sessionId.winner"] = winner
+      }
       // Perform targeted update in Firestore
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/android/streetworkapp/model/workout/WorkoutRepositoryFirestore.kt
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * Updates specific details of a WorkoutSession.
+   *
+   * @param uid The UID of the user (also the WorkoutData ID).
+   * @param sessionId The ID of the session to update.
+   * @param exercises A list of exercises to add or update.
+   * @param endTime The end time of the session.
+   */
+  override suspend fun updateWorkoutSessionDetails(
```
kzepfl, line 70: Why only allowing to update the list of exercises and endTime ?
I think we would at least to be able to update the winner aswell as we only know the winner at the end of the session.


```
    val startTime: Long,
    val endTime: Long,
    val sessionType: SessionType,
    val participants: List<String> = emptyList(),
    val exercises: List<Exercise> = emptyList(),
    val winner: String? = null
```

app/src/test/java/com/android/streetworkapp/model/workout/WorkoutRepositoryFirestoreTest.kt
```diff
@@ -0,0 +1,182 @@
...
+    verify(documentRef)
+        .update(
+            mapOf(
+                "workoutSessions.$sessionId.exercises" to exercises,
+                "workoutSessions.$sessionId.endTime" to endTime))
+  }
+
+  @Test
+  fun updateExerciseUpdatesCorrectFieldPath() = runTest {
+    val uid = "testUid"
```
kzepfl, line 161: There are many val and when that appear multiple time in every test, I think some of them can be put in `setup`. Like  val uid = "testUid" for example

kzepfl (approved):
Appart from `updateWorkoutSessionDetails` that could need some change the PR is approved

app/src/main/java/com/android/streetworkapp/model/workout/WorkoutRepositoryFirestore.kt
```diff
@@ -0,0 +1,195 @@
...
+              ?.map(::parseWorkoutSession) ?: emptyList()
+      WorkoutData(userUid, workoutSessions)
+    } catch (e: Exception) {
+      Log.e(ERROR_TAG, "Error converting document to WorkoutData: ${e.message}")
+      WorkoutData("Error check logcat", emptyList())
+    }
+  }
+
+  /**
+   * Converts a Firestore document to a WorkoutSession object.
```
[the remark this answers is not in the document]
tercierp: I did rename this

app/src/main/java/com/android/streetworkapp/model/workout/WorkoutRepositoryFirestore.kt
tercierp: Yep changes done

PR #136, opened 2024-11-20
Feat/training UI

alvaro1080:
**Overview**
In order to allow the user to track his progression in the different exercices, this Pull Request adds a dashboard with a "Training" tab in the Progression UI. It will be used to see the last record of the user on a specific exercise. Its purpose is also to navigate to a screen with all the details about this exercise, that should be implemented in the future.

This feature have strong connections with the "train with a friend" feature. Some changes will be needed in the progression MVVM to keep track of the records in the different exercises, and it will be the objective of a future PR. 

**Changes**
- Add a Dashboard in the Progression Screen, to either see the achievement or the record of each exercise
- Define a new enum class to represent the exercises, and use it in the UI.
- Update the corresponding tests, with the use of this Dashboard

**How to test**

1. go to the Progression Screen
2. Click on the "Training" or the "Achievement" part of the Dashboard

you should be able to see your achievement/the record of your training!

(Note: the button are not enable, because the detail screen does not exist yet. This is the reason why the color is not the same as the other button. Once enable, the color should be the same as the rest of the UI)

**Screenshot**
image


app/src/main/res/drawable/handstand.png
SaturneV, line 1: I think it would be a good idea to change the background color of this icon so that it's the same as the other icon. Is it possible to do this?

SaturneV: Or the opposite, of course.

app/src/main/java/com/android/streetworkapp/model/progression/Progression.kt
```diff
@@ -33,7 +33,7 @@ data class Achievement(
     val icon: Int,
     val title: String,
     val tags: List<String>,
-    val description: String
+    var description: String
```
SaturneV, line 36: Why does it have to be a variable? Can the description change after initialization?

app/src/main/java/com/android/streetworkapp/model/progression/Progression.kt
```diff
@@ -69,6 +69,11 @@ enum class MedalsAchievement(val achievement: Achievement) {
           "Last medal"))
 }
 
+enum class ExerciseAchievement(val achievement: Achievement) {
```
SaturneV, line 72: I see that it hasn't been done in the other `enum class` statements in this file, but it would be a good idea to put a short description of what this class represents.

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
```diff
@@ -145,31 +148,43 @@ fun ProgressScreen(
...
+                    }
+                  }
+                }
+              }
+              DashboardStateProgression.Training -> {
+
+                enumValues<ExerciseAchievement>().forEach {
+                  it.achievement.description =
+                      "Record : 0.0 sec" // TODO: use the progression MVVM with current record
+                  Box(modifier = Modifier.testTag("exerciseItem" + it.name)) {
```
SaturneV, line 183: I personally find `"exerciseItem${it.name}"` cleaner, but that's just a suggestion, fell free to leave it like that.

SaturneV (changes requested):
Great PR! The description is well detailed, the UI is UX friendly and in compliance with our design on Figma, and the tests cover the new implementations well. I'll let you answer my comments. Good job!

app/src/main/java/com/android/streetworkapp/model/progression/Progression.kt
alvaro1080: Yes, it is used in the exercises achievements in the UI, to display the record of the user. We hope that the user will do a better record, and thus we will need to update the description.

app/src/main/res/drawable/handstand.png
alvaro1080: I am not an expert in design, but I tried to make the 2 background the same. I updated the pictures, tell me if it is better.

app/src/main/java/com/android/streetworkapp/model/progression/Progression.kt
alvaro1080: I think you are right. I added a description for every enum in this class.

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
alvaro1080: Done, thank you!

app/src/main/res/drawable/handstand.png
SaturneV: Perfect, thank you!

app/src/main/java/com/android/streetworkapp/model/progression/Progression.kt
SaturneV: But doesn't updating the ViewModel overwrite the current progress with a new object every time an update is made in the database, making the variable concept useless?

I'll let you have a look at the User ViewModel for example, the User data class has only values and not variables - every repository operation update the `_user` value with a new value.

What do you think of it? It's a detail, so if you don't want to change, no problem, but did you understand what I was talking about?

alvaro1080: I think you are confused between this enum and the attributes of the ViewModel. A new attribute will probably be added, similar to the _currentUser you are talking about, with a mutableStateFlow etc... 
Here, this enum is not a mutable state flow, it is juste a way to store information about Achievements. (Notice that the progression collection in the  database does not store any achievement, only a "name" of this achievement, that is then used to display the correct element of the enum)
If it is still not clear, we can discuss it in person as it is quite complicated by messages.

SaturneV: We talked about it in person - I hadn't quite understood how achievements worked, but now it's fine and we'll leave it that way.

SaturneV (approved):
My comments have been answered. I approve this PR, it can be merge.

PR #148, opened 2024-11-23
Feat/implement helping dialogs in screens

Simmanz:
## Description

The aim of this PR is to add helping pop-ups for each important screens, this was done using a new action `INFO` and defining classes to do so. 

I started by re-working the `CustomDialog` to handle multiple types of pop-ups. Then, I created new components to handle the dialogs in a cleaner way, `InfoDialog` and `InfoDialogManager`.

### CustomDialog improvement

- Now possible to define the `title` that is displayed.
- Now changed the old `dialogType` to `tag`, because it was about the instance / use of the Dialog, not its realy type (`SETTINGS`, `INFO`, ... )
- There is now multiple **types** : `QUERY` and `INFO`. The first is the usual, while `INFO` does not have buttons / submit logic.

### New Action (IconButton)

- New topAppBarAction : `INFO` that is used for the help screens.

### Implementation in `MainActivity`

- State `showInfoDialog` that handles info pop-up being displayed or not.
- `InfoDialog` class, which is a specific version of CustomDialog with type `INFO`.
- `InfoDialogManager` class, which handles the content to be displayed in the dialogs depending on the current screen (fetched from the navigation).
- `setUp()` and `Display()` functions from the manager handle the callbacks and display respectively.
### Strings.xml

I prepared the file with all the "hardcoded" values, the problem is that to extract such values, we are forced to use context. Which totally broke the current implementation in main. Therefore the link with strings will be finished later.

### Tests 

- Updated the `CustomDialogTest` to verify the new behavior.
- Updated the `ParkRatingTest` that uses the `CustomDialog` for the Rating.
- Created `InfoDialogsTest` which verifies that everything is displayed correctly + manager behaves correctly.
## Testing the features

- Click on the '**?**' action in the screens.
- A pop-up should be displayed with basic information about the screen.
- Pop-up should be closed by clicking outside of it.

<p align="center">
image
</p>


app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
```diff
@@ -0,0 +1,137 @@
...
+ */
+class InfoDialogManager(
+    private val showDialog: MutableState<Boolean>,
+    private val screenName: MutableState<String?>,
+    private val topAppBarManager: TopAppBarManager?
+) {
+  private val screens = getScreens()
+
+  // Mapping Screen to InfoDialog
+  private val infoDialogs: MutableMap<String, InfoDialog> =
```
SaturneV, line 28: I think it would be a good idea to add a `TODO` here to explain that this must be implemented in the future via `strings.xml`. I would even open an issue explaining the problem in order to keep track of the problem. What do you think of it?

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
```diff
@@ -0,0 +1,137 @@
...
+                  "This is the progression screen, here you can see your progression and achievements. \n\n There is your record for each exercise in Training, click on it to see more details !")
+
+          // Auth not done because unnecessary
+          // Add more if needed
+          )
+
+  /** Returns the default InfoDialog (for screens that are not supported). */
+  fun defaultInfoDialog(): InfoDialog {
+    return InfoDialog(
+        "Default",
```
SaturneV, line 74: Personally, I'd use the title **"Screen not implemented"** instead of **"Default"**, to make it more user-friendly. But that's just a suggestion, keep it that way if you prefer!

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
```diff
@@ -0,0 +1,137 @@
...
+   * This function displays the info dialog, which is a specific Type of CustomDialog.
+   *
+   * @param showDialog - The display state of the dialog.
+   */
+  @Composable
+  fun DisplayInfoDialog(showDialog: MutableState<Boolean>) {
+    CustomDialog(
+        showDialog,
+        dialogType = DialogType.INFO,
+        tag + "Info",
```
SaturneV, line 127: I personally find `"${tag}Info"` cleaner, but that's just a suggestion, fell free to leave it _and the all the others I have commented_ like that.

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
```diff
@@ -0,0 +1,137 @@
...
+        title,
+        Content = { DisplayInfoContent(content) })
+  }
+
+  /** Function wrapper to allow more customization later on */
+  @Composable
+  fun DisplayInfoContent(content: String) {
+    Text(content, modifier = Modifier.testTag(tag + "InfoDialogContent"))
```
SaturneV, line 135: Same here, I find `"${tag}InfoDialogContent"` cleaner.

app/src/main/java/com/android/streetworkapp/ui/utils/CutsomDialog.kt
```diff
@@ -35,37 +39,41 @@ fun CustomDialog(
 
   if (showDialog.value) {
     AlertDialog(
-        modifier = Modifier.testTag(dialogType + "Dialog"),
+        modifier = Modifier.testTag(tag + "Dialog"),
```
SaturneV, line 42: Same here, I find `"${tag}Dialog"` cleaner.

app/src/main/java/com/android/streetworkapp/ui/utils/CutsomDialog.kt
```diff
@@ -35,37 +39,41 @@ fun CustomDialog(
...
-                Text("Submit", color = ColorPalette.SECONDARY_TEXT_COLOR)
-              }
+          if (dialogType.shouldShowSubmitButton()) {
+            TextButton(
+                onClick = {
+                  onSubmit()
+                  Toast.makeText(context, "Submitted", Toast.LENGTH_SHORT).show()
+                  showDialog.value = false
+                },
+                modifier = Modifier.testTag(tag + "DialogSubmitButton")) {
```
SaturneV, line 55: Same here, I find `"${tag}DialogSubmitButton"` cleaner.

app/src/main/java/com/android/streetworkapp/ui/utils/CutsomDialog.kt
```diff
@@ -35,37 +39,41 @@ fun CustomDialog(
...
+                }
+          }
         },
         title = {
           Text(
-              "Your $dialogType",
+              title,
               color = ColorPalette.PRIMARY_TEXT_COLOR,
-              modifier = Modifier.testTag(dialogType + "DialogTitle"))
+              modifier = Modifier.testTag(tag + "DialogTitle"))
```
SaturneV, line 76: Same here, I find `"${tag}DialogTitle"` cleaner.

app/src/main/java/com/android/streetworkapp/ui/utils/CutsomDialog.kt
```diff
@@ -74,3 +82,16 @@ fun CustomDialog(
...
+  QUERY, // Have submit / cancel buttons + use OnSubmit / onDismiss.
+  INFO; // Only display the content + use onDismiss.
+
+  fun shouldShowSubmitButton(): Boolean {
+    return this == QUERY
+  }
+
+  fun shouldShowCancelButton(): Boolean {
+    return this == QUERY
+  }
```
SaturneV, line 96: Just one quick question - what is the purpose of these two functions and why not just use the enum types?

SaturneV (changes requested):
Great PR! The new feature is a great addition to the application and I especially want to thank you for the very detailed PR description(!) and the quality of your architecture(!). I'll let you have a look at my few comments, which are just details. Good job!

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
Simmanz: I agree with you, it would ensure that we do not forget about the refactoring !

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
Simmanz: Actually the first argument is the tag, which is only used for testTags purposes (or when no title is defined).

In this case the title is "Screen is not supported yet".

You can check this by commenting-out one element of the mapping and navigating there. That would make the call to the map return null, and the display will give the default parameters.

app/src/main/java/com/android/streetworkapp/ui/utils/CutsomDialog.kt
Simmanz: I did this because we might want more types of Dialogs in the future, I was thinking ahead and tried to get as much of the logic out of the composable as possible. (If there are 4-5 types it would quickly get convoluted in the Dialog's code)

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
Simmanz: I agree, I will change it right now.

misterM125 (changes requested):
Great PR, the idea of implementing pop ups to help the user familiarize himself with the different feature of the app is a fantastic idea and i like the way you implemented it.
I mainly brought a few typos that i think could be corrected as well as a one suggestion to make the code a bit more clean

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
```diff
@@ -0,0 +1,138 @@
...
+      Log.d("InfoDialog", "Displaying info dialog : ${func.title}")
+      func.DisplayInfoDialog(showDialog)
+    } else {
+      Log.d("InfoDialog", "Displaying default info dialog : ${defaultInfoDialog().title}")
+      defaultInfoDialog().DisplayInfoDialog(showDialog)
+    }
+  }
+
+  /** Returns the InfoDialogs mapping */
+  fun getInfoDialogs(): MutableMap<String, InfoDialog> {
```
misterM125, line 111: Suggestion: as a convention, it is always better to use Kotlin implicit getters instead of creating your own getter function:
```
// Mapping Screen to InfoDialog
  private val _infoDialogs: MutableMap<String, InfoDialog> = ...
    val infoDialogs: MutableMap<String, InfoDialog> 
        get() = _infoDialogs
```

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
```diff
@@ -0,0 +1,137 @@
...
+          screens.ADD_EVENT to
+              InfoDialog(
+                  "AddEvent",
+                  "How to create an event ?",
+                  "This is your firsts steps towards creating an event ! \n\n You simply have to add a cool title, then, describe your event : What are you going to do there ? \n\n Finally, set the date and time and you are good to go !"),
+          screens.EVENT_OVERVIEW to
+              InfoDialog(
+                  "EventOverview",
+                  "What is the Event Overview ?",
+                  "This is the event overview, here you can access all of the information that you need. \n\n You are free to join the event if you want to participate !"),
```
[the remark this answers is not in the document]
Simmanz: I do not completely agree because it is inside the Event Overview; the user must have clicked on a specific event. I think that it is better to specify the event in this place, whereas in parkoverview it would totally make sense.

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
```diff
@@ -0,0 +1,137 @@
...
+                  "This is your profile page, you can see your friends and your score. \n\n You can also add friends using the 'Add friend' button and you can also access your settings by clicking the gear icon on the top right."),
+          screens.ADD_FRIEND to
+              InfoDialog(
+                  "AddFriend",
+                  "How do friends work ?",
+                  "This is where you can add a friend. \n\n To add, you need to have a friend nearby and activate bluetooth and location services. \n\n Once you are friends you will be able to do workouts with each other. \n\n Have fun together !"),
+          screens.PARK_OVERVIEW to
+              InfoDialog(
+                  "ParkOverview",
+                  "How does parks work ?",
```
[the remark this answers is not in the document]
Simmanz: I agree, I missused the _do vs does_ syntax.

app/src/main/java/com/android/streetworkapp/ui/utils/CutsomDialog.kt
```diff
@@ -14,19 +14,23 @@ import androidx.compose.ui.window.DialogProperties
 import com.android.streetworkapp.ui.theme.ColorPalette
```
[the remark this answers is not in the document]
Simmanz: Thanks for catching the typo !

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
SaturneV: Thank you for adding the TODO! Please just open an Issue with the **Refactor** tag so we can track this problem in the future.

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
SaturneV: You're right, thanks for the clarification!

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
Simmanz: You can access it here : https://github.com/SwEnt-Group8/Street-work-app/issues/169

app/src/main/java/com/android/streetworkapp/ui/utils/CutsomDialog.kt
Simmanz: This whole PR is done with scalability in mind, for this reason I think it is better as-is for these functions. As we seem to disagree, I propose to refactor them as `shouldShowButtons()`, which is self-explanatory about its use. Do you agree with this proposition ?

misterM125: Agreed, those are functions that are easy to implement later if necessary, furthermore scalability or not two functions doing the same thing is code duplication so as long as refactor them into one this is a good a change

app/src/main/java/com/android/streetworkapp/ui/navigation/InfoDialogs.kt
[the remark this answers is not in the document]
misterM125: actually you are right, i misread and did not fully pay attention to the context in which this message was used
Apologies

misterM125 (approved):
Good changes, this PR is now good to merge for me!

SaturneV (approved):
Thank you for the changes and responses to my comments. I approve this PR. Good job!

PR #149, opened 2024-11-23
Feat/integrate text moderation in UI

Alex720p:
This PR aims to integrate the text moderation in UI, make the addEvent design responsive and fix any existing/new issues.

- **`MainActivity.kt`**  
  - Defined the TextModeration repo & vm and passed them as parameters accordingly.

- **`PerspectiveAPIRepository.kt`**  
  - Fixed issue where app would crash on calling analyzeText function from UI by having it being run in DispatcherIO.

- **`AddEventScreen.kt`**  
  - Changed the position of the button in the date selection for it to be more responsive.
  - Applied padding from the `Scaffold` to the `AddEventScreen` composable for consistent layout spacing.  
  - Replaced the `FloatingActionButton` with a `Button` to allow for more responsive UI.  
  -  Removed height specification on OutlinedTextField to allow them to resize correctly on other screen sizes
  - Made the `AddEventScreen` screen scrollable to make it usable on different screen sizes.  
  - Added persistent form error messages for user input validation (e.g., required fields and text moderation). 
    - The messages dissapears as soon as the user performs an input in the fields where the error is located.
  - Displayed transient error messages (visible for 5 seconds) for non-user-input-related errors (API/network issues).  
  - Fixed a bug where the time in the event model reset to `00:00` if the form couldn't be submitted due to an error. (Note: this was **not** a bug that existed before since the screen didn't recompose) 
  - Fixed issue where the `event` model wasn't synced with the DatePicker on initialization.  
  - Fixed issue in `convertMillisToDate` where timezone was hardcoded to utc instead of choosing the system one.
  - Fixed issue where time calculation for `currentTimeSelection` was wrong and added an offset of one hour to the timestamp.
  - Added a `*` for required fields and added annotation for it at bottom of form.

- **`DateUtils.kt`**
  - Removed the hardcoded UTC timezone to the system default

- **`Test Updates`**  
  - Updated existing tests to support the changes in the `AddEventScreen` screen and MainActivity.
  - Moved event-related UI test files into the `ui` package for better organization. 
  - Removed `timeSelectionUpdatesEvent` from `AddEventTest.kt` as the init time is now synced to the event and thus the test is not compatible anymore
  - Implemented new tests for the form error messages

Here's a few screenshots showcasing the changes, all the error messages are shown below:
<div style="display: flex; justify-content: center;">

</div>


app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -103,57 +133,115 @@ fun AddEventScreen(
...
+                          } else {
+                            isTextEvaluationOverThresholdsError.value = true
+                            formErrorMessage =
+                                AddEventFormErrorMessages.TEXT_EVALUATION_OVER_THRESHOLDS_ERROR
+                          }
+                        },
+                        {
+                          isTextEvaluationError = true
+                          formErrorMessage = AddEventFormErrorMessages.TEXT_EVALUATION_ERROR
+                        })
```
misterM125, line 240: For the same reasons as the above suggestion, i would suggest separating this code block (from 214 to 240) into a separate function to reduce complexity and avoid nested ifs, best way i managed to achieve this is by changing the definition of `formErrorMessage` to make it a `MutableState`:
`val formErrorMessage = remember { mutableStateOf("") }`

and modifying line 213 up to 241 like so:
```
if (formErrorMessage.value.isEmpty()) {
                  val error =
                      analyzeTextAndCreateEvent(
                          event,
                          navigationActions,
                          eventViewModel,
                          userViewModel,
                          parkViewModel,
                          context,
                          isTextEvaluationOverThresholdsError,
                          textModerationViewModel,
                          formErrorMessage)
                  isTextEvaluationError = error
                }
```

and create this private function:
```
private fun analyzeTextAndCreateEvent(
    event: Event,
    navigationActions: NavigationActions,
    eventViewModel: EventViewModel,
    userViewModel: UserViewModel,
    parkViewModel: ParkViewModel,
    context: Context,
    isTextEvaluationOverThresholdsError: MutableState<Boolean>,
    textModerationViewModel: TextModerationViewModel,
    formErrorMessage: MutableState<String>
): Boolean {
  val formattedTextInput = "Title: ${event.title}. Description: ${event.description}"
  var isTextEvaluationError = false
  textModerationViewModel.analyzeText(
      formattedTextInput,
      { isTextUnderThresholds ->
        if (isTextUnderThresholds) {
          createEvent(
              event, navigationActions, eventViewModel, userViewModel, parkViewModel, context)
        } else {
          isTextEvaluationOverThresholdsError.value = true
          formErrorMessage.value = AddEventFormErrorMessages.TEXT_EVALUATION_OVER_THRESHOLDS_ERROR
        }
      },
      {
        isTextEvaluationError = true
        formErrorMessage.value = AddEventFormErrorMessages.TEXT_EVALUATION_ERROR
      })

  return isTextEvaluationError
}
```

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIRepository.kt
```diff
@@ -56,56 +58,55 @@ class PerspectiveAPIRepository(private val client: OkHttpClient) : TextModeratio
    * @return TextEvaluationResult.Success if the API could process the content,
    *   TextEvaluationResult.Error if an error was encountered
    */
-  private fun getTextAnnotations(content: String): PerspectiveAPIEvaluationResult {
-    // I don't perform any checks on the content, I'll let the api handle the error checking
+  private suspend fun getTextAnnotations(content: String): PerspectiveAPIEvaluationResult {
+    return withContext(Dispatchers.IO) {
```
misterM125, line 62: Suggestion: testing might be hindered by using a hardcoded dispatcher, i would recommend using `Dispatchers.Default` instead

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIRepository.kt
```diff
@@ -22,7 +24,7 @@ class PerspectiveAPIRepository(private val client: OkHttpClient) : TextModeratio
    * @return True if the text is under all thresholds, false if the text is over at least one
```
misterM125, line 24: Suggestion: i would recommend using a companion object to store DEBUG_PREFIX as a constant (Line 18): 
```
companion object {
        private const val DEBUG_PREFIX = "PerspectiveAPIRepository:"
    }
```
(I cannot comment unchanged lines but it seemed relevant for this PR)

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIRepository.kt
Alex720p: Why would the testing being hindered? My reasoning for the choice of `Dispatchers.IO` is that the operation we're doing is more network heavy than anything else. The api call should be the thing that takes the most amount of time here.

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIRepository.kt
Alex720p: done

app/src/main/java/com/android/streetworkapp/model/moderation/TextModerationViewModel.kt
```diff
@@ -14,20 +14,20 @@ open class TextModerationViewModel(val repository: TextModerationRepository) : V
    *   successfully evaluated. The value passed as parameter is true if the text is under all
    *   thresholds, false otherwise
    * @param onTextEvaluationError A callback function to be executed if an error occurs during the
-   *   request. The error message is passed as parameter
+   *   request.
    * @param thresholds A map of threshold values for different text moderation tags. If not
    *   provided, default values are used.
    */
-  fun analyzeText(
+  open fun analyzeText(
```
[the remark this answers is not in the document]
Alex720p: Can't mock functions that aren't open in mockito

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -242,32 +339,41 @@ fun ParticipantNumberSelection(event: Event) {
...
-  val currentTime = Calendar.getInstance()
-  val datePickerState = rememberDatePickerState(currentTime.timeInMillis)
+  val initialTime = Calendar.getInstance().apply { timeInMillis = event.date.toDate().time }
+  val datePickerState = rememberDatePickerState(initialTime.timeInMillis)
   val timePickerState =
       rememberTimePickerState(
-          initialHour = currentTime.get(Calendar.HOUR_OF_DAY),
-          initialMinute = currentTime.get(Calendar.MINUTE),
+          initialHour = initialTime.get(Calendar.HOUR_OF_DAY),
+          initialMinute = initialTime.get(Calendar.MINUTE),
```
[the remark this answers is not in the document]
Alex720p: done

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
Alex720p: I refactored the code in a function called `onAddEventClickHandler`

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
```diff
@@ -103,57 +133,115 @@ fun AddEventScreen(
...
+                        contentColor = ColorPalette.PRIMARY_TEXT_COLOR),
+                modifier = Modifier.testTag("addEventButton"),
+                onClick = {
+                  if (event.date.seconds <
+                      Timestamp(Calendar.getInstance(TimeZone.getDefault()).time).seconds) {
+                    isDateBackInTimeError.value = true
+                    formErrorMessage = AddEventFormErrorMessages.DATE_BACK_IN_TIME_ERROR
+                  } else if (event.title.isEmpty()) {
+                    isTitleEmptyError.value = true
+                    formErrorMessage = AddEventFormErrorMessages.TITLE_EMPTY_ERROR_MESSAGE
```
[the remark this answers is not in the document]
Alex720p: Honestly I like the current flow better. I feel like it's much more readable

app/src/main/java/com/android/streetworkapp/model/moderation/PerspectiveAPIRepository.kt
misterM125: This is something that was noted by sonarcloud mainly, injectable dispatchers ease testing by allowing tests to inject more deterministic dispatchers. But i guess this is a problem that we can deal with when we actually encounter it, you can do as you want for now

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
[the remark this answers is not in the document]
misterM125: I can understand that, up to you

app/src/main/java/com/android/streetworkapp/ui/event/AddEvent.kt
misterM125: Nice! this seems to have greatly reduced the cognitive complexity, thank you for the changes

misterM125 (approved):
Thank you for the changes, i understand you concerns regarding some of the changes i suggested, up to you if you want to keep them or not

PR #152, opened 2024-11-24
Feat/Update and set gamification in relevant parts of the UI

alvaro1080:
## Overview
The objective of this Pull Request is to implement the gamification feature in the relevant part of the app. To do so, a `Snackbar` is displayed whenever the user wins points, with the amount of points obtained. This `Snackbar` allows the user to go see directly the progression screen updated with the new score by clicking on the "see Progression" part of the `Snackbar.`
An increase of score and a display of the `Snackbar` is added in three parts of the app in this Pull Request: when the user creates an event, when he joins an event and when he adds a new friend.

## Changes
- A new function `updateAndDisplayPoints` is created and used in order to increase the score of the user and display the `Snackbar`
- Creating an event, joining an event and adding a friend now adds points
- Details of the Progression UI screen are also updated (add missing icon for medals and update achievements description)

## How to test
1. Go to the Map Screen
2. Select any park, and add or join an event

You should see a `Snackbar` appear, with the option to "See Progression" (see screenshot if it is not clear), you can click on the "See Progression" part, which will redirect you to the Progression Screen.

3. Check that your score increased

## Screenshot
image


app/src/main/java/com/android/streetworkapp/ui/progress/Gamification.kt
```diff
@@ -0,0 +1,36 @@
...
+) {
+
+  userViewModel.currentUser.value?.let { userViewModel.increaseUserScore(it.uid, points) }
+  scope.launch {
+    val result =
+        snackbarHostState.showSnackbar(
+            message = "You obtained $points Points!",
+            actionLabel = "See Progression",
+            withDismissAction = true,
+            duration = SnackbarDuration.Indefinite)
```
Alex720p, line 27: Suggestion: having to click on the dismiss to close it is annoying. I think it would be better to set a non indefinite delay. (5 seconds maybe?)

app/src/test/java/com/android/streetworkapp/model/progression/ProgressionModelTest.kt
```diff
@@ -0,0 +1,14 @@
+package com.android.streetworkapp.model.progression
+
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+
+class ProgressionModelTest {
+
+  @Test
+  fun getOrAddProgressionCallsRepository() = runTest {
```
Alex720p, line 9: this is pointless, if this is what you needed to get over the 80% coverage, we can just remove the rule and merge your branch instead. If we modify the point system, it will be annoying

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
```diff
@@ -39,16 +40,23 @@ import com.android.sample.R
...
 import com.android.streetworkapp.ui.theme.ColorPalette.PRIMARY_TEXT_COLOR
 import com.android.streetworkapp.ui.theme.ColorPalette.PRINCIPLE_BACKGROUND_COLOR
 import com.android.streetworkapp.ui.theme.ColorPalette.SECONDARY_TEXT_COLOR
 import com.android.streetworkapp.ui.theme.LightGray
+import kotlinx.coroutines.CoroutineScope
 
 @Composable
 fun AddFriendScreen(
     userViewModel: UserViewModel,
+    navigationActions: NavigationActions? = null,
```
Alex720p, line 57: Why is it a nullable var? navigationActions will always be defined

app/src/main/java/com/android/streetworkapp/model/user/UserViewModel.kt
```diff
@@ -158,7 +158,14 @@ open class UserViewModel(private val repository: UserRepository) : ViewModel() {
    * @param points The number of points to add to the user's score.
    */
   fun increaseUserScore(uid: String, points: Int) =
-      viewModelScope.launch { repository.increaseUserScore(uid, points) }
+      viewModelScope.launch {
```
Alex720p, line 161: It's dangerous that we need to update the user manually here. Would be better to return the user from repo.increaseUserScore and just store it inside currentUser instead

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
```diff
@@ -88,7 +96,12 @@ fun AddFriendScreen(
         username = user.username,
         onAccept = {
           userViewModel.addFriend(uid, receivedRequestUid)
-          Toast.makeText(context, "Friend added.", Toast.LENGTH_SHORT).show()
+          // Toast.makeText(context, "Friend added.", Toast.LENGTH_SHORT).show()
+          if (host != null && navigationActions != null) {
```
Alex720p, line 100: no need check if navigationsActions is non-null, it's always the case when passed from MainActivity

Alex720p (changes requested):
The snackbar works great!
I left a few comments in the code below but it's mostly details, everything that has to do with functionnality works great.

I do have a few comments about the design of the snackbar:
- Having a black color like this really doesn't blend well with our app design. 
I think that having the background color close to the white background we have with setting the 'You obtained ...' and the cross to the font color of the figma would blench much better with the rest of the app.
- The infinite duration of the snackbar is annoying, putting it to a non-indifinite time would be much better.

app/src/main/java/com/android/streetworkapp/ui/progress/Gamification.kt
alvaro1080: You are right, I update it, thank you!

app/src/test/java/com/android/streetworkapp/model/progression/ProgressionModelTest.kt
alvaro1080: The intention of the test is unclear, I apologize. The objective is to make sure that the hierarchy of the points is respected. The rationale is that we want user to join an existing event rather than creating one themself, and that adding a friend should be the action with the most points, as this is the main objective of the app. The test as been updated to be more clear, if you want to check.

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
alvaro1080: You are right, this is updated thank you!

app/src/main/java/com/android/streetworkapp/ui/profile/AddFriend.kt
alvaro1080: This as been updated, thank you!

app/src/main/java/com/android/streetworkapp/model/user/UserViewModel.kt
alvaro1080: I understand your concern. I think that the rework of the repo is outside the scope of this PR, we could try to make another version in the upcoming sprint.

Alex720p: yeah but it should be resolved in another pr:
since we have
`        _currentUser.value?.let {
          val updatedUser = it.copy(score = it.score + points)
          _currentUser.value = updatedUser
        }`

in the viewmodel that gets executed even if firebase fails.
Someone could: turn off internet, spam an action that increases points, turn on internet again, go to progression and have checkscore being triggered. Then he will receive all the achievements even if he doesn't have the points.

Alex720p:
Note: We're not sure about the color of the snackbar. It will be asked during the standup and have a fix later on if we decide to change it.

Alex720p (approved):

PR #160, opened 2024-11-26
Feat/Get and set current user

SaturneV:
### Summary 

The purpose of this short PR is to implement a new function in the User ViewModel that gets a user by user id and sets it as the current user, enabling this operation to be performed in a single line and thus avoiding synchronous problems when this task was performed in two operations. 

### Changes 

- Add the `getUserByUidAndSetAsCurrentUser` in `UserViewModel.kt`

### Testing

- Implement tests to test `getUserByUidAndSetAsCurrentUser` in `UserViewModelTest.kt`

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/model/user/UserRepository.kt
+++ b/app/src/main/java/com/android/streetworkapp/model/user/UserRepository.kt
@@ -6,6 +6,8 @@ interface UserRepository {
 
   suspend fun getUserByUid(uid: String): User?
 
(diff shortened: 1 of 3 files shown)


alvaro1080 (approved):
Thank you for this pull request! This new function `getUserByUidAndSetAsCurrentUser` is correctly tested and documented, since this Pull Request only focuses on this addition, I do not have a lot more to say. I approve this Pull Request.

PR #161, opened 2024-11-27
hotfix/event overview ui

misterM125:
**Overview**
This PR aims to solve all issues relating to the event overview screen as well as finish implementing multiple event related features

**implemented features**
- The event overview screen now displays correctly the name of the event owner instead of its uid
- the event map screen now displays the correct marker location of the event's park
- the park overview screen now displays relative time for each event instead of a specific date
- The event overview screen will now also properly display the list of all participants to the related event
- slight changes to the event overview ui have been made to fill up empty space

**Additionnal info**
- fixed issues with the `documentToPark` function that did not properly convert `Parklocation` parameters
- fixed issues where the number of participants to an event would sometimes differ from the number of participants displayed in the list of participants
- tests related to user, park, and event mvvm as well as the event and park ui have been refactored to take the aforemntioned changes into account

**Testing**
These changes can be tested by:
- creating an event, you should now be able to see in how many days / hours your event is going to take place
- then if you go to the event overview screen you should see your name displayed as well as your profile picture in the event details
- The event map screen will show a marker displaying the location of the park where your event is taking place, clicking on said marker should redirect you to the park overview screen for that park
- Ask friends to come join your event on their end! ( if you don't have any friends you can always manually add users via the firestore database by copying their uid into the "ParticipantList" field of your event.
- the list of participants of your event (accessible by clicking on the "participants" tab of the event dashboard) will then display their username with their profile picture


app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
```diff
@@ -161,15 +168,23 @@ fun EventMap(park: Park) {
...
       }
 
   GoogleMap(
       modifier = Modifier.testTag("googleMap").fillMaxSize(),
       cameraPositionState = cameraPositionState) {
         Marker(
-            contentDescription = "Marker",
-            state = MarkerState(position = LatLng(parkLatLng.latitude, parkLatLng.longitude)))
+            contentDescription = "Event location",
+            state = MarkerState(position = parkLatLng ?: LatLng(0.0, 0.0)),
```
alvaro1080, line 179: Suggestion: Since now the markers are blue on the main map, they could also be changed to blue on the event overview in order to make the UI coherent.

app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
```diff
@@ -321,7 +321,12 @@ class ParkRepositoryFirestore(private val db: FirebaseFirestore, testing: Boolea
     return try {
       val pid = document.id
       val name = document["name"] as? String ?: ""
-      val location = document["location"] as? ParkLocation ?: ParkLocation(0.0, 0.0, "0")
+      val locationMap = document["location"] as? Map<*, *> ?: emptyMap<String, Any>()
```
alvaro1080, line 324: Question: I think I am missing something: what is the difference between this version of the code and the last one?

app/src/androidTest/java/com/android/streetworkapp/ui/event/EventOverviewTest.kt
```diff
@@ -65,6 +69,25 @@ class EventOverviewTest {
     screenParams = LIST_OF_SCREENS.last()
 
     navigationActions = mock(NavigationActions::class.java)
+    // fullevent = event.copy(participants = 10, maxParticipants = 10)
```
alvaro1080, line 72: Comment: This commented part of code could perhaps be removed.

alvaro1080 (changes requested):
Thank you for this pull request! This improvement of the event overview screen is great. Being able to see the profile picture of the participants is a nice improvement, and knowing in how much time an event take place is more user friendly than just a date.

I have a few suggestions that I would like to discuss:

app/src/main/java/com/android/streetworkapp/ui/event/EventOverview.kt
misterM125: You a right, i'll change it

app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
misterM125: Essentially, the previous version is just incorrect: parklocations are stored as `Map` values in the firestore database so trying to naively cast it to ParkLocation when fetching from the database results in a null value (since what's fetched is not a `Parklocation`) so i end up with parks that just have the default `ParkLocation(0.0, 0.0, "0")`

This is the reason the markers were not properly displayed in the event overview map to begin with
Thank you for your inquiry and let me know if this is not clear enough!

app/src/androidTest/java/com/android/streetworkapp/ui/event/EventOverviewTest.kt
misterM125: done!

alvaro1080 (approved):
Thank you for the updates!

PR #163, opened 2024-11-27
Change maps markers colors to be coherent with the rest of the UI

alvaro1080:
## Overview
The objective of this Pull Request is to change the default markers we currently have in the map, in order to obtain a more coherent UI in the app.

(Note: This Pull Request was originally supposed to implement more meaningful feature to the map, but I blocked and I could not finish it. To make sure that we have a coherent UI for the next demo, I propose the current Pull Request instead.)

## Changes

- The color of the markers are now coherent with the rest of the app.

## How to test
Once you go to the map screen, the markers should now appear as blue.

## Screenshot

image


tercierp (approved):
Very good change, it really adds to the visual experience. Approved !

PR #180, opened 2024-12-02
Linking infodialogs with strings.xml file to avoid hard-coded strings

Simmanz:
## Assistants gave a solution that works 

- Apply R.strings (**Type** `Int` instead of `String`)
- Then, in **@Composable** `display()` functions, use `Context.getString(Int)`

### Some changes in main

I choose to use `LocalContext.current` for each call to the `InfoDialogManager.display()` instead of using one context variable. I did this because the screens are not the same, which implies changes to the scaffold / other elements, which might break the **@Composable** if I use an older context.

### Changing tests

Because the type of the `InfoDialog` changed, needed to re-write some tests which used specific strings.

### There should be no changes feature-wise

<p align="center">
image
</p>


kzepfl (approved):
Good PR, it worked on my side.

PR #182, opened 2024-12-02
fix(Tutorial): make dot appear

kzepfl:
image

## Changes

- Change in `TutorialEvent.kt` and `TutorialSignIn.kt` added back the sizes


Simmanz (approved):
### Fast fixing !
Its great that you were fast in fixing the issue since I notified you about it ! Good job !

The dots are now correctly being displayed again, you also did remember to make it change depending on which page you are on : 

<p align="center">
image
image
</p>

Issue #144, opened 2024-11-22
Finish implementing the settings for the profile


Simmanz:
### Acceptance criteria : 

- The Setting dialog in the profile has some settings available for the user.

### Settings : 

- [ ] Log out 
- [ ] Account Deletion
- [ ] Privacy with localisation (proximity devices, allow localisation)

PR #185, opened 2024-12-02
Feat/implement friends deletion through per friend parameter

Simmanz:
## Description

This PR aims to enable users to remove their friends through the `FriendMenu`, which is accessible through the "..." in the `FriendList` (Profile screen).

### Implementation

- Created new type for **CustomDialog** (pop-up) : `CONFIRM` - which is simple display with **confirm button**.
- Added a **menu for actions** on friend : `FriendMenu`
- Added `RemoveFriendMenuItem` accessible through the **menu**. Which allow user to remove a friend.
- Clicking the remove friend action will open a confirmation dialog **before removing the friend**.

/!\ The friend list is now updated after removing a friend ! (Call to `getFriendsByUid`)

### Strings.xml
The confirmation dialog's title and text content are stored in `strings.xml` and extracted in the menu.

### New implementation since feedback
Added new handlers to prevent cognitive complexity with the current button display (if, else if). 
It also clearly improves the code.

#### `HandleConfirmButton`

Refactoring of the if-logic for which button to display (`SUBMIT`, `CONFRIM`).

#### `HandleDismissButton`

Refactoring of the if-logic for which button to display (`CANCEL`).

#### `DialogButton`

Handles the display of any button.

- `ButtonType` - the type of the button (used to extract the text values, testTag, and color).
- `tag: String` - the tag of the dialog (used for testTag).
- `onSubmit: () -> Unit` - callback of the button.
-` showDialog: MutableState<Boolean>` - used to hide the dialog after clicking.
- `verbose: Boolean` - whether the Toast() should be called (refactoring).

#### New functions for `enum class ButtonType`

- `fun response(context): String` - The text for click confirmation (Toast).
- `fun tag(context): String` - The tag for testTag and title of the button.
- `fun color(): Color` - The color of the button.

Note : both reponse and tag now give their `String` values from **strings.xml**.

### Tests

- Updating tests for `Profile` and its **Components** (`FriendList`, `FriendListItem`) in `ProfileComponentsTest`
- Created new test that verify friend removal functionality in `RemoveFriendTest`.
- Created other tests to check ButtonHandlers functionality

### Testing the functionality

1. By clicking on the "**...**" on the right of a **friend** in the **friend list**, a menu should **appear**.
2. The menu should have a "**remove friend**" option.
3. Clicking on the remove friend should open a pop-up asking for confirmation.
4. After confirming, you should no longer have the friend in your friend list (try re-loading the page).

<p align="center">
image
image
image
</p>


app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -167,14 +177,16 @@ fun DisplayFriendList(friends: List<User?>) {
 }
 
 /**
- * This function displays a friend (for the friend list).
+ * This function displays a friend (for the friend list). Also displays actions to perform on them
+ * through a dropdown menu.
```
SaturneV, line 181: To make the documentation easier to read, I'd write it in a single sentence:

 “_This function displays a friend (for the friend list) and actions to perform on them through a dropdown menu._”.

It's only a suggestion, I leave it to you to choose what you prefer.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -258,3 +274,79 @@ fun DisplayUserPicture(user: User?, size: Dp, testTag: String) {
...
+            modifier = Modifier.testTag("RemoveFriendMenuItem"),
+            onClick = {
+              val friendUID = friend.uid
+              val friendName = friend.username
+
+              val currentUID = userViewModel.currentUser.value?.uid
+
+              if (currentUID != null) {
+                // Set up the confirm function for the dialog
+                Log.d("Profile", "Setting up callback (confirm) for removing friend")
```
SaturneV, line 321: It would be a good idea to store the “_Profile_” log tag in a const variable and use it in the rest of the file. What do you think of it?

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
```diff
@@ -258,3 +274,79 @@ fun DisplayUserPicture(user: User?, size: Dp, testTag: String) {
...
+            },
+            text = {
+              // Display content here
+              Row() {
+                Icon(
+                    painter = painterResource(id = R.drawable.person_remove),
+                    contentDescription = "Remove friend",
+                    modifier = Modifier.size(24.dp),
+                    tint = Color.Red)
+                Text("Remove friend", modifier = Modifier.padding(start = 4.dp), color = Color.Red)
```
SaturneV, line 348: Personally, I'd put this hard coded text in the `strings.xml` file. In case we want to implement the application in several languages, it's a text we'd like to be able to change easily.

SaturneV: I also notice that this is the case for Toast, but I think that here it's more important and that you can leave the hard coded string of Toast - Toast is not part of the primary understanding of the app.

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
```diff
@@ -85,9 +97,20 @@ fun CustomDialog(
 
 enum class DialogType {
   QUERY, // Have submit / cancel buttons + use OnSubmit / onDismiss.
-  INFO; // Only display the content + use onDismiss.
+  INFO, // Only display the content + use onDismiss.
+  CONFIRM;
```
SaturneV, line 101: It would be a good idea to add a comment like for the enums above.

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
```diff
@@ -45,23 +46,34 @@ fun CustomDialog(
           showDialog.value = false
         },
         confirmButton = {
```
SaturneV, line 48: Sonar raise an issue about the cognitive complexity of this `CustomDialog` composable function - and I agree with it. To reduce cognitive complexity, you could just implement functions that correspond to the action you want to perform. So it's just a matter of copying and pasting your existing lines into functions and using those functions in the composable. 

Here's an example, instead of having
```kotlin
confirmButton = {
if (dialogType.shouldShowButton()) {
          if (dialogType.shouldShowButton(ButtonType.SUBMIT)) {
            TextButton(
                onClick = {
                  onSubmit()
                  Toast.makeText(context, "Submitted", Toast.LENGTH_SHORT).show()
                  if (verbose) Toast.makeText(context, "Submitted", Toast.LENGTH_SHORT).show()
                  showDialog.value = false
                },
                modifier = Modifier.testTag("${tag}DialogSubmitButton")) {
                  Text("Submit", color = ColorPalette.SECONDARY_TEXT_COLOR)
                }
          } else if (dialogType.shouldShowButton(ButtonType.CONFIRM)) {
            TextButton(
                onClick = {
                  onSubmit()
                  if (verbose) Toast.makeText(context, "Confirmed", Toast.LENGTH_SHORT).show()
                  showDialog.value = false
                },
                modifier = Modifier.testTag("${tag}DialogConfirmButton")) {
                  Text("Confirm", color = Color.Red)
                }
          }
        }
```

You can juste take theses lines, put them in a `handleOnConfirm(...)` function and use it in the composable

```kotlin
confirmButton = handleOnConfirm(...)
```

It should pretty quick, what do you think of it?

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
```diff
@@ -33,7 +33,8 @@ fun CustomDialog(
     title: String = "Your $tag",
     Content: @Composable () -> Unit = {},
     onSubmit: () -> Unit = {},
-    onDismiss: () -> Unit = {}
+    onDismiss: () -> Unit = {},
+    verbose: Boolean = true
```
SaturneV, line 37: Here Sonar has also raised an issue about too many function parameters. At first sight, it's not really possible to reduce the number of parameters. What do you think, there's room for improvement or not?

SaturneV (changes requested):
Great PR! Once again a very good architecture and understanding of the elements and also a very detailed PR description. I'll let you answer my comments. Just a suggestion - to make the description cleaner, don't hesitate in the future to use tags to make the images smaller and add descriptions to them (see #144). Good job!

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: I agree that this could be improved. I changed it to :

"_This function displays a friend (of the friend list) and actions about them through a dropdown menu._"

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: I agree that it can prevents typos, which would lead to logs not appearing / multiple profile tags.

app/src/main/java/com/android/streetworkapp/ui/profile/Profile.kt
Simmanz: Thanks for catching that one ! I did not see it. I will define both of them while I am at it :)

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
Simmanz: Thanks again for catching this, I forgot to change it afterwards !

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
Simmanz: You are totally right, I will define a handler for the buttons.

Simmanz: ## New implementation

### `HandleConfirmButton`

Refactoring of the if-logic for which button to display (`SUBMIT`, `CONFRIM`).

### `HandleDismissButton`

Refactoring of the if-logic for which button to display (`CANCEL`).

### `DialogButton`

Handles the display of any button.

- ButtonType - the type of the button (used to extract the text values, testTag, and color).
- tag: String - the tag of the dialog (used for testTag).
- onSubmit: () -> Unit - callback of the button.
- showDialog: MutableState<Boolean> - used to hide the dialog after clicking.
- verbose: Boolean - whether the Toast() should be called (refactoring).

### New functions for `enum class ButtonType`

- fun response(): String - The text for click confirmation (Toast).
- fun tag(): String - The tag for testTag and title of the button.
- fun color(): Color - The color of the button.

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
Simmanz: I understand your point, I do agree with the fact that it is not really reducible : the principle of the `Dialog` is to be able to fit any pop-up need. 

Therefore, removing `onDismiss` or `onSubmit` would lessen the value of such a class for the codebase. This is also why I put default functions to simplify the usage as much as possible.

Also removing the `Content` function would prevent us from customizing the current pop-ups as we want such as **the rating of parks**, which uses a specific component to display interactible stars to rate, not simply text.

Simmanz:
>Just a suggestion - to make the description cleaner, don't hesitate in the future to use tags to make the images smaller and add descriptions to them (see https://github.com/SwEnt-Group8/Street-work-app/issues/144). Good job!

I believe I am not quite sure what you are talking about ? #144 is the issue I made for the settings in the profile and contains no image so far.

I already use a fixed size for images, as well as putting them in a paragraph using **\<p>**, do you mean to make them even smaller ?

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
SaturneV: What a detailed answer! Thank you for making the changes. It's great like this.

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
```diff
@@ -84,10 +63,106 @@ fun CustomDialog(
...
+ *
+ * @param ButtonType - Type of button (SUBMIT, CANCEL, CONFIRM)
+ * @param tag - Describe the instance of the dialog ("Settings", "Rating", etc.)
+ * @param onSubmit - Submission callback
+ * @param showDialog - MutableState to show the dialog
+ * @param verbose - Whether to display a Toast message on interaction
+ */
+@Composable
+fun DialogButton(
+    ButtonType: ButtonType,
```
SaturneV, line 121: Sonar raised an issue about the parameter naming, please do not use a capital letter for the first letter.

SaturneV (changes requested):
Thanks for replying to my comments and once again great work on the modifications! I still let you fix one last parameter naming issue and also accept the too many parameters issue on Sonar. Good job!

app/src/main/java/com/android/streetworkapp/ui/utils/CustomDialog.kt
Simmanz: The changes have been pushed

SaturneV (approved):

PR #187, opened 2024-12-03
Feat/park searching in map and marker info windows

misterM125:
test


Alex720p (changes requested):
Good pr, the markers are really clean.
I've commented on the naming of the testtags to follow our naming convention, nothing to say about the code otherwise.


**UI related comments:**
- The icons in the top app bar aren't aligned, even if the difference is slight it looks weird in the app.  
icons not aligned_review

- The difference in thickness between the 2 icons is also a bit weird, might be worth changing the logo to something thicker (this for ex: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:map_search:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=search&icon.size=24&icon.color=%23e8eaed) but it's up to you I'll approve if there's no change.

- For the searchbar you should fix the padding so that it isn't exactly at the bottom but vertically centered. The color the searchbar uses isn't in the figma, it looks a bit out of place, but again up to you to change this or not.
seachbar_review

**Bugs:**
There is a bug in the search, reading your implementation I really don't think it's coming from your end, but it should be fixed in another pr. Same for the duplicate parks that mess up the UX with the markers, but also not from your end (issue created for this: https://github.com/SwEnt-Group8/Street-work-app/issues/189).
Here's an example: I search for Route de L... and the park doesn't show up when it should
review_search_before_dissapear
review_search_after_dissapear
This doesn't happen everytime.

Alex720p:
Looks good! Thanks for the changes

PR #195, opened 2024-12-05
Feat/finish implementing the settings for the profile

Simmanz:
## Description

This PR aims to implement the UI for the first settings of the app : 

- **Logout Button** - whose logic is not implemented in this PR.
- **Delete account Button** - Which calls a Dialog (pop-up) that asks for confirmation before exectuing a callback function. Real deletion is not implemented in this PR.

### Test
New test has been made in `SettingsTest`.

### Testing the feature

- In the profile, clicking on the gear icon should open the settings.
- Clicking on "delete account" should open a confirmation pop-up.

<p align="center">
image
image
image
</p>

### DIFF
--- a/app/src/androidTest/java/com/android/streetworkapp/ui/profile/SettingsTest.kt
+++ b/app/src/androidTest/java/com/android/streetworkapp/ui/profile/SettingsTest.kt
@@ -1,0 +1,92 @@
+package com.android.streetworkapp.ui.profile
+
+import android.content.Context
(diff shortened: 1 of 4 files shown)


kzepfl (approved):
Good PR, it worked on my PC.
My only nitpick would be the button "Confirm", maybe a border would be useful for the user to know it's a button.

Simmanz:
The buttons are those of an AlertDialog, which is used in the CustomDialog. I will see what we can do about that last sprint.

PR #201, opened 2024-12-07
Setup image backend & early draft for the park image collection composable

Alex720p:
This pr implements the backend necessary for image uploading, integrating it into our UI as well as an early (non finished) draft for the image collection view of a park.

**General Overview:**
In the db Parks now contain a new field `imagesCollectionId`, by default set to empty string. When a user uploads the first picture to the park, a collection is created and its id is linked to the park. Inside the collection, the images are stored in a list with each item in this format:
`
    val imageB64: String,
    val userId: String,
    val rating: Pair<Int, Int>,
    val uploadDate: Timestamp
`

**Backend**
**`ParkRepositoryFirestore:`**
Added function `addImagesCollection` allowing us to links images collections to parks.

**`ImageRepositoryFirestore`:**
Allows us to upload images in base64 to firestore, this allows us to keep using the existing db and not having to setup a payment card for the storage in firebase :). We can then retrieve the images from the db.
The process could be greatly optimized but for our use case, the current implementation will work fine.

**`ImageViewModel`:**
Wrapper for the repository. For the retrieving, it saves the images in the app's cache.

**Ui**
**`ImagesCollectionButton`:**
A new button displayed in `ParkOverview` next to the `AddImageButton`. Clicking it will open the collection popup (very early draft and will be worked on in a later pr), allowing the user to view the park's images.
**`FullScreenImagePopup`:**
The collection popup that displays the park's images. The user can swipe to see the different images.

**Known bugs**
On a park with no `imagesCollectionId` setup yet, taking the first picture and going to collection will not show the images. This is due to the park not updating its field after a call to firebase. Will fix it in later pr by implementing a listener or something. For now, exiting the screen and clicking on the park again will refresh the park data and display the image.

**Pictures of the changes**
Here's the button in `ParkOverview` (bottom right of the park image) & the collection popup:
<div style="display: flex;">
  
  
</div>


app/src/main/java/com/android/streetworkapp/model/image/ImageRepositoryFirestore.kt
```diff
@@ -0,0 +1,115 @@
...
+  override suspend fun retrieveImages(park: Park): List<ParkImageDatabase> {
+    require(this.parkRepository.getParkByPid(park.pid) != null) { "Invalid ParkId." }
+
+    if (park.imagesCollectionId.isEmpty())
+        return emptyList() // no collection setup yet, just return no images
+
+    try {
+      // checking the collection linked to the park is valid
+      val document = db.collection(COLLECTION_PATH).document(park.imagesCollectionId).get().await()
+      if (!document.exists())
```
alvaro1080, line 80: Refactor: this could be refactored using "require()"

app/src/main/java/com/android/streetworkapp/model/image/ImageRepositoryFirestore.kt
```diff
@@ -0,0 +1,115 @@
...
+          throw IllegalArgumentException(
+              "The ImagesCollection linked by this park doesn't exist, the entry in the db is corrupted.")
+
+      val parkImages = document.get("images") as? List<*>
```
alvaro1080, line 84: Refactor: this "get" could be changed to a [] (similar to what was used in line 91 for instance)

app/src/main/java/com/android/streetworkapp/ui/image/ConfirmImageDialog.kt
```diff
@@ -164,3 +171,45 @@ fun ConfirmImageDialog(
...
+  }
+}
+
+// TODO: make UI responsive depending on success/failure
+private fun onImageUploadSuccess(file: File) {
+  if (!file.delete())
+      Log.d(AddImageButtonParams.DEBUG_PREFIX, "Failed to delete cached photo file.")
+}
+
+private fun onImageUploadFailure() {}
```
alvaro1080, line 215: Perhaps we could add a small comment inside this empty {}

app/src/main/res/drawable/photo_library_24px.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="960"
+    android:viewportHeight="960">
+    <!--android:tint="?attr/colorControlNormal">-->
+  <path
+      android:fillColor="@android:color/white"
+      android:pathData="M360,560L760,560L622,380L530,500L468,420L360,560ZM320,720Q287,720 263.5,696.5Q240,673 240,640L240,160Q240,127 263.5,103.5Q287,80 320,80L800,80Q833,80 856.5,103.5Q880,127 880,160L880,640Q880,673 856.5,696.5Q833,720 800,720L320,720ZM320,640L800,640Q800,640 800,640Q800,640 800,640L800,160Q800,160 800,160Q800,160 800,160L320,160Q320,160 320,160Q320,160 320,160L320,640Q320,640 320,640Q320,640 320,640ZM160,880Q127,880 103.5,856.5Q80,833 80,800L80,240L160,240L160,800Q160,800 160,800Q160,800 160,800L720,800L720,880L160,880ZM320,160L320,160Q320,160 320,160Q320,160 320,160L320,640Q320,640 320,640Q320,640 320,640L320,640Q320,640 320,640Q320,640 320,640L320,160Q320,160 320,160Q320,160 320,160Z"/>
+</vector>
```
alvaro1080, line 10: Question: I am not sure I understand why this file is used for?

app/src/main/java/com/android/streetworkapp/model/image/ImageModels.kt
```diff
@@ -0,0 +1,52 @@
...
+// ImageRepositoryFirestore::uploadImage
+
+/**
+ * A data class that represents an image related to a park in the database, uploaded by a user, and
+ * associated with a rating.
+ *
+ * @property imageB64 The image in Base64 string format.
+ * @property userId The id of the user who uploaded the image.
+ * @property rating A pair of integers representing the rating, where the first element represents
+ *   all the positive ratings and the second the negative ones.
```
alvaro1080, line 25: Question: I am not sure I understand why the property rating in contains "the first element represents
 all the positive ratings and the second the negative ones", could you explain this choice?

app/src/main/java/com/android/streetworkapp/model/image/ImageModels.kt
```diff
@@ -0,0 +1,52 @@
...
+ * A data class that represents an image related to a park in the app, uploaded by a user, and
+ * associated with a rating.
+ *
+ * @property image The [Uri] of the image.
+ * @property userId The id of the user who uploaded the image.
+ * @property rating A pair of integers representing the rating, where the first element represents
+ *   all the positive ratings and the second the negative ones.
+ * @property uploadDate The date the image was uploaded
+ */
+data class ParkImageLocal(
```
alvaro1080, line 47: Question: Is the separation between ParkImageLocal and ParkImageDatabase necessary? (for instance, could not we just have a ParkImage, that we sent to the database and we use locally?

alvaro1080 (changes requested):
Thank you for your Pull Request! It is an important step for the park picture feature. The code is clean, I only have a few questions and a few details that I would like to discuss:

app/src/main/java/com/android/streetworkapp/model/image/ImageModels.kt
Alex720p: yep I think so. In the db you store the imageb64, but in the app you retrieve it and store them in the cache that you keep track via uri's. It doesn't make sense to keep long imageb64 in memory to decode it on the fly, decoding them once and storing them is better. Also this allows to skip the decoding if the image is already in the cache.

app/src/main/java/com/android/streetworkapp/model/image/ImageModels.kt
Alex720p: basically it's Pair(#the number of ppl who liked this picture, #the number of ppl who disliked this picture)

app/src/main/res/drawable/photo_library_24px.xml
Alex720p: It's the icon for the button to access the collection.

app/src/main/java/com/android/streetworkapp/ui/image/ConfirmImageDialog.kt
Alex720p: done

app/src/main/java/com/android/streetworkapp/model/image/ImageRepositoryFirestore.kt
Alex720p: done

app/src/main/java/com/android/streetworkapp/model/image/ImageRepositoryFirestore.kt
Alex720p: done

app/src/main/java/com/android/streetworkapp/model/image/ImageModels.kt
alvaro1080: I now see better were it is going, thank you!

app/src/main/java/com/android/streetworkapp/model/image/ImageModels.kt
alvaro1080: I see, thank you for the explanation!

alvaro1080 (approved):
Thank you for the answers and the updates!

Issue #171, opened 2024-11-29
Implement the account deletion ability


SaturneV:
### Description

This feature will implement the ability to delete an account i.e. delete completely the account from the device and from the database. 

### Tasks

- Implement an account deletion function.
- Implement this functionality in the UI.
- Test this new feature.

PR #204, opened 2024-12-08
Feat/Logout

SaturneV:
### Summary

This PR implements the ability to logout from an account on the application. This involves logging out of the Google Auth Service, clearing users from the user viewmodel and clearing preferences parameters. 

In order to implement this feature, it was necessary to implement a function enabling a user to be set in the `_user` value of the user viewmodel.

The initialization of the Google Auth Service was moved to the main activity to enable its use in both the SignIn and Settings features. Additionally, the Google Auth Service class was refactored to improve usability across the project. These updates also pave the way for implementing account deletion functionality #171.

### Changes

- Implement a `SetUser` function in `UserViewModel.kt`.
- Instantiate the Google Auth Service in `MainActivity.kt`.
- Refactor the `GoogleAuthService` class.
- Implement a logout function in `Settings.kt`.
- Link the logout function with the UI.
- Add the logout related string in the ressources.

### Tests

- Modify existing tests by adding a Google Auth Service mock.
- Implement a `SetUser` test function in `UserViewModelTest.kt`.
- Implement a new UI test in `SettingsTest.kt`.

**Please fully test this feature with your device / emulator to make sure it works properly.**


app/src/main/java/com/android/streetworkapp/ui/profile/Settings.kt
```diff
@@ -87,3 +96,29 @@ fun SettingsContent(
...
+  // Clear user viewmodel data
+  userViewModel.setUser(null)
+  userViewModel.setCurrentUser(null)
+
+  // Clear preferences parameters
+  preferencesViewModel.setLoginState(false)
+  preferencesViewModel.setUid("")
+  preferencesViewModel.setName("")
+  preferencesViewModel.setScore(0)
+}
```
tercierp, line 124: Improvement: The data is removed inside the PreferencesViewModel, but when I try to re-login using the Google button, I am automatically logged in with my previous account. 
After some research, I found that the Google access token needs to be revoked to fully log out the user. 
This can be done using the following method:
```kotlin
authService.revokeAccess()
```
I may have done something wrong, but it's strange that I can't choose which google account to log in with. What do you think ?

tercierp (changes requested):
Very clear code ! But there's probably a problem when the user and delog something retains with which account I previously logged. And when I relog it logs directly with the previous account without offering me a choice.

app/src/main/java/com/android/streetworkapp/ui/profile/Settings.kt
SaturneV: Good catch and thanks for the video, it helped me understand the problem. In fact, the signout function of the `GoogleAuthService` class didn't take care of properly resetting the client, so it wasn't properly disconnected and therefore it wasn't possible to choose a Google account to log in. I've made the changes.

I'll let you test if it works on your side. Let me know if you have any other questions or suggestions.

tercierp:
Yes, it works now. Thank you for adding this important addon.

PR #206, opened 2024-12-09
Feat/ Create achievements for training with the app

alvaro1080:
## Overview
The objective of this Pull Request is to give to the user achievements when the training part of the app is used. With the social achievements and the medals, the progression screen now rewards all the aspect of the app.

The new achievements could be improved with the "detail" button that is currently removed. It will come back with a new associated screen in a future PR, in order to keep the size of the current PR as small as possible.

## Changes
- Add the workout ViewModel in the Progression Screen
- New achievements, base on the workout ViewModel
- The corresponding tests are updated for the new kind of achievement

## How to test

You can use the "solo training" feature of the app to train. Then, a new achievement should appear in the Progression screen, similar to the one in the screenshot below.

## Screenshot

image

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
+++ b/app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
@@ -54,8 +54,11 @@ import com.android.streetworkapp.model.progression.MedalsAchievement
 import com.android.streetworkapp.model.progression.ProgressionViewModel
 import com.android.streetworkapp.model.progression.SocialAchievement
 import com.android.streetworkapp.model.user.UserViewModel
+import com.android.streetworkapp.model.workout.WorkoutRepositoryFirestore
+import com.android.streetworkapp.model.workout.WorkoutViewModel
 import com.android.streetworkapp.ui.navigation.NavigationActions
 import com.android.streetworkapp.ui.theme.ColorPalette
+import com.google.firebase.firestore.FirebaseFirestore
 import kotlinx.coroutines.flow.MutableStateFlow
 
 object ProgressionScreenSettings {
@@ -76,6 +79,8 @@ fun ProgressScreen(
     navigationActions: NavigationActions, // Note: not used yet
     userViewModel: UserViewModel,
     progressionViewModel: ProgressionViewModel,
+    workoutViewModel: WorkoutViewModel =
+        WorkoutViewModel(WorkoutRepositoryFirestore(FirebaseFirestore.getInstance())),
     paddingValues: PaddingValues = PaddingValues(0.dp)
 ) {
 
--- a/app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
+++ b/app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
@@ -49,7 +49,6 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.android.sample.R
 import com.android.streetworkapp.model.progression.Achievement
-import com.android.streetworkapp.model.progression.ExerciseAchievement
 import com.android.streetworkapp.model.progression.MedalsAchievement
 import com.android.streetworkapp.model.progression.ProgressionViewModel
 import com.android.streetworkapp.model.progression.SocialAchievement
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
```diff
@@ -76,12 +79,17 @@ fun ProgressScreen(
     navigationActions: NavigationActions, // Note: not used yet
     userViewModel: UserViewModel,
     progressionViewModel: ProgressionViewModel,
+    workoutViewModel: WorkoutViewModel =
+        WorkoutViewModel(WorkoutRepositoryFirestore(FirebaseFirestore.getInstance())),
```
SaturneV, line 83: I'm not sure it's a good idea to leave the possibility of implicitly declaring the viewmodel here, it should be done in the MainActivity to avoid multiple instances. Do you agree with me?

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
```diff
@@ -165,12 +173,20 @@ fun ProgressScreen(
...
-                  Box(modifier = Modifier.testTag("exerciseItem${it.name}")) {
-                    AchievementItem(it.achievement, true)
-                    HorizontalDivider(thickness = 0.5.dp, color = Color.LightGray)
+                currentWorkout?.workoutSessions?.forEach { workout ->
+                  val date = workout.startTime.toFormattedString()
+                  workout.exercises.forEach {
+                    val newAchievement =
+                        Achievement(
+                            exerciseNameToIcon(it.name),
+                            "You trained your ${it.name} !",
```
SaturneV, line 182: Avoid using hard-coded strings and use resource strings instead.

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
```diff
@@ -370,3 +386,18 @@ sealed class DashboardStateProgression {
...
+    "Push-ups" -> R.drawable.train_pushup
+    "Dips" -> R.drawable.train_dips
+    "Burpee" -> R.drawable.train_burpee
+    "Lunge" -> R.drawable.train_lunge
+    "Planks" -> R.drawable.train_planks
+    "Handstand" -> R.drawable.train_hand_stand
+    "Front lever" -> R.drawable.train_front_lever
+    "Flag" -> R.drawable.train_flag
+    "Muscle-up" -> R.drawable.train_muscle_up
+    else -> R.drawable.handstand_org
```
SaturneV, line 401: Here too, instead of using hard coded strings just use the resource string to do the mapping.

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
```diff
@@ -370,3 +386,18 @@ sealed class DashboardStateProgression {
 
   data object Achievement : DashboardStateProgression()
 }
+
+fun exerciseNameToIcon(name: String): Int {
```
SaturneV, line 390: The function is quite explicit, but a definition of the function and its parameters would be useful.

SaturneV (changes requested):
Great PR, I'll let you reply to my few comments and also remind you that we use resource strings now following the coaches recommendations. Good job!

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
alvaro1080: You are right, I removed it and I updated the tests.

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
alvaro1080: Done, thank you!

app/src/main/java/com/android/streetworkapp/ui/progress/Progression.kt
alvaro1080: Done, thank you!

SaturneV (approved):
Thanks for the changes and for responding to my comments. It's good for me, I approve this PR. Good job!

PR #216, opened 2024-12-11
Fix/e2e create event

misterM125:
## Overview

This PR aims to refactor the end to end to create an event by removing all network calls and mocking the database to make it valid

### Implementation
- This end to end now tests the full functionalities of all repositories and viewmodels called when navigating from the map to a park by clicking on the park overview, as well as creating an event and displaying it on the park overiew

- The end to end had to be divided into two parts, I suspect that the CI has some limitations that i am not aware of but the test won't pass on it if it is too long, leaving me only with the choice of making it significantly smaller or dividing it into two parts.

### Additionnal info
- Markers have been refactored so that you navigate to the park overview screen only by clicking on the marker info window, this is a change that had been requested and i decided to implement it here because it actually made testing marker functionalities easier
- A test in eventOverviewTest has been modified because it wouldn't pass on the CI anymore after writing my end to end (?), similar functionalities are tested in the end to end so i decided to reduce the scope of  that test to make it pass


app/src/main/java/com/android/streetworkapp/model/park/ParkViewModel.kt
```diff
@@ -40,6 +40,10 @@ open class ParkViewModel(
     _currentPark.value = park
   }
```
tercierp, line 42: Improvement: since `setCurrentPark` has a doc could you also add a small function description ?

tercierp (approved):
Very good work, I already approve your PR because I only saw a cleaning code issue and a function that needs to be documented.

PR #219, opened 2024-12-12
Setup online file storage system

Alex720p:
Very simple S3 wrapper. This allows us to interact with the digital ocean space bucket to save the images from our app.
Note: This needs to get out of the way pretty fast for me to continue on my other pr. There are no tests. 
This resolves the issue of not being able to store large images with the free plan of firebase.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -41,6 +41,9 @@ android {
     //fetching keys
     val mapsApiKey: String = localProperties.getProperty("MAPS_API_KEY") ?: ""
     val perspectiveApiKey: String = System.getenv("PERSPECTIVE_API_KEY") ?: localProperties.getProperty("PERSPECTIVE_API_KEY", "")
+    val digitalOceanSpaceAccessKey: String = System.getenv("DIGITAL_OCEAN_SPACE_ACCESS_KEY") ?: localProperties.getProperty("DIGITAL_OCEAN_SPACE_ACCESS_KEY", "")
+    val digitalOceanSpaceSecretKey: String = System.getenv("DIGITAL_OCEAN_SPACE_SECRET_KEY") ?: localProperties.getProperty("DIGITAL_OCEAN_SPACE_SECRET_KEY", "")
+
 
     defaultConfig {
         applicationId = "com.android.streetworkapp"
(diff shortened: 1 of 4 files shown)


app/build.gradle.kts
```diff
@@ -56,6 +59,9 @@ android {
 
         manifestPlaceholders["MAPS_API_KEY"] = mapsApiKey
         buildConfigField("String", "PERSPECTIVE_API_KEY", "\"${perspectiveApiKey}\"")
+        buildConfigField("String", "DIGITAL_OCEAN_SPACE_ACCESS_KEY", "\"${digitalOceanSpaceAccessKey}\"")
+        buildConfigField("String", "DIGITAL_OCEAN_SPACE_SECRET_KEY", "\"${digitalOceanSpaceSecretKey}\"")
+
```
tercierp, line 64: Question: Did you check that the builder APK takes the correct secrets?

tercierp (changes requested):
I put several questions, because I'm not sure if it's normal or not but for the moment I can't test anything, is that right?

app/src/main/java/com/android/streetworkapp/model/storage/S3StorageClient.kt
```diff
@@ -0,0 +1,100 @@
...
+
+enum class S3Clients(
+    val endpoint: String,
+    val region: String,
+    val accessKey: String,
+    val secretKey: String
+) {
+  DIGITAL_OCEAN(
+      endpoint = "https://street-work-app.fra1.digitaloceanspaces.com",
+      region = "us-east-1",
```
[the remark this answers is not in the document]
Alex720p: the region doesn't actually matter for digitalocean

> Due to an AWS-specific behavior in all versions of the SDK except Python 3, to successfully create a new bucket, you must specify an AWS region, such as us-east-1, in your configuration. This is because, when creating a bucket, the SDK sends an entirely different payload if a custom region is specified, which results in an error.
Specifying us-east-1 does not result in slower performance, regardless of your bucket’s location. The SDK checks the region for verification purposes but never sends the payload there. Instead, it sends the payload to the specified custom endpoint.

https://docs.digitalocean.com/products/spaces/how-to/use-aws-sdks/

app/build.gradle.kts
Alex720p: nope, was thinking of doing this in the upcoming pr. Would be easier to test as the pictures would be implemented there.

Alex720p:
> I put several questions, because I'm not sure if it's normal or not but for the moment I can't test anything, is that right?

Yep, nothing to test and nothing you can actually do to test as you don't have the keys yet. Will give them Friday on team meeting.

tercierp (approved):

tercierp:
Thank you for your precision approved

PR #221, opened 2024-12-12
Feat/Delete account

SaturneV:
### Summary

This Pull Request implements the ability to delete an account. This feature is implemented in the setting screen of the application. It required the modification of several MVVM in order to remove data from the database, as well as the deletion of the Firebase Google Authentication user.
 
When deleting an account, it is possible to have to reconnect before deleting, so it has been necessary to be able to launch the login composable from the settings. 

### Changes

- Add a global friend deletion function in User MVVM.
- Add a global event participation deletion function in Event MVVM.
- Add a global rating deletion function in Park MVVM.
- Add a progression deletion function in Progression MVVM.
- Add a workout deletion function in Workout MVVM.
- Implement a user deletion function in `GoogleAuthService.kt`.
- Implement a `deleteAccount` function in `Settings.kt`
- Integrate the account deletion function in the `Settings.kt` UI.

### Tests

- Add several MVVM tests.
- Modify existing tests to add new parameters.

**Please fully test this feature with your device / emulator to make sure it works properly.**


app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
```diff
@@ -311,6 +311,30 @@ class ParkRepositoryFirestore(private val db: FirebaseFirestore, testing: Boolea
     }
   }
```
misterM125, line 313: Suggestion: you should also implement a function to delete the deleted user's events from a park

app/src/test/java/com/android/streetworkapp/ui/profile/SettingsTest.kt
```diff
@@ -0,0 +1,116 @@
...
+    verify(userViewModel, never()).removeUserFromAllFriendsLists(anyString())
+    verify(parkViewModel, never()).deleteRatingFromAllParks(anyString())
+    verify(eventViewModel, never()).removeParticipantFromAllEvents(anyString())
+    verify(progressionViewModel, never()).deleteProgressionByUid(anyString())
+    verify(workoutViewModel, never()).deleteWorkoutDataByUid(anyString())
+  }
+
+  @Test
+  fun testDeleteAccountFailsDueToNullFirebaseUser() {
+    whenever(userViewModel.currentUser).thenReturn(MutableStateFlow(currentUser))
```
misterM125, line 95: This call could be put in the setUp instead of repeating it in every test

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
```diff
@@ -156,6 +156,42 @@ class EventRepositoryFirestore(private val db: FirebaseFirestore) : EventReposit
...
+   * Remove a participant from all events and delete events where the user is the owner.
+   *
+   * @param uid The user ID.
+   */
+  override suspend fun removeParticipantFromAllEvents(uid: String) {
+    require(uid.isNotEmpty())
+    try {
+      val events =
+          db.collection(COLLECTION_PATH)
+              .whereArrayContains(FIELD_LIST_PARTICIPANTS, uid)
```
misterM125, line 169: Suggestion: this is more of an edge case but it might problematic if a uid contains the same letters as another. To prevent that, you can use `whereEqualTo` instead

misterM125 (changes requested):
Great PR! I have tested the functionalities to delete an account on my end on the emulator and they do seem to be working as intended.

I do have a couple suggestions to avoid code duplication and improve readbility
I also think it would be useful to also implement a function to delete a deleted owner's event from a parkList

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
SaturneV: The `whereArrayContainsAny()` function definition is the following:
> Creates and returns a new Query with the additional filter that documents must contain the specified field, the value must be an array, and that the array must contain the provided value.

This function check for arrays (in our case the participants list) that contains a certain value (in our case the participant to remove) and the check is made on the entire value according to its documentation. In this case I think that my implementation is correct, what do you think of it?

app/src/main/java/com/android/streetworkapp/model/user/UserRepositoryFirestore.kt
```diff
@@ -295,6 +295,33 @@ class UserRepositoryFirestore(private val db: FirebaseFirestore) : UserRepositor
...
+  /**
+   * Removes a user from all friends lists in Firestore.
+   *
+   * @param uid The unique ID of the user to remove from all friends lists.
+   */
+  override suspend fun removeUserFromAllFriendsLists(uid: String) {
+    require(uid.isNotEmpty()) { UID_EMPTY }
+    try {
+      // Get all users that have the user in their friends list
+      val querySnapshot = db.collection("users").whereArrayContains("friends", uid).get().await()
```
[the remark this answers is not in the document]
SaturneV: Good suggestion, I've made the suggested change.

app/src/main/java/com/android/streetworkapp/model/workout/WorkoutRepositoryFirestore.kt
```diff
@@ -251,6 +251,24 @@ class WorkoutRepositoryFirestore(private val db: FirebaseFirestore) : WorkoutRep
...
 
+  /**
+   * Deletes the entire WorkoutData for a user from Firestore.
+   *
+   * @param uid The UID of the user.
+   */
+  override suspend fun deleteWorkoutDataByUid(uid: String) {
+    require(uid.isNotEmpty()) { ERROR_UID_EMPTY }
+    try {
+      val querySnapshot = db.collection(COLLECTION_PATH).whereEqualTo("userUid", uid).get().await()
```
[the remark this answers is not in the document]
SaturneV: Here too.

app/src/test/java/com/android/streetworkapp/ui/profile/SettingsTest.kt
SaturneV: Good suggestion, I've made the suggested change.

app/src/main/java/com/android/streetworkapp/model/event/EventRepositoryFirestore.kt
misterM125: You are right, i misread what the function was actually doing, my apologies

app/src/main/java/com/android/streetworkapp/model/park/ParkRepositoryFirestore.kt
SaturneV: Good catch, I have made this change! It consists of saving the list of parks deleted when using the event viewmodel and then passing this list to the park viewmodel in order to delete them from all parks. I'll let you have a look at this addition and let me know what you think of it.

app/src/main/java/com/android/streetworkapp/model/user/UserRepositoryFirestore.kt
```diff
@@ -295,6 +295,33 @@ class UserRepositoryFirestore(private val db: FirebaseFirestore) : UserRepositor
     }
   }
 
+  /**
+   * Removes a user from all friends lists in Firestore.
+   *
+   * @param uid The unique ID of the user to remove from all friends lists.
+   */
+  override suspend fun removeUserFromAllFriendsLists(uid: String) {
```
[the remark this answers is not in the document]
SaturneV: This function is now tested!

app/src/androidTest/java/com/android/streetworkapp/ui/profile/SettingsTest.kt
```diff
@@ -130,4 +173,80 @@ class SettingsTest {
...
+  fun testDeleteAccountButtonNavigatesToAuthScreen() {
+    val showSettingDialog = mutableStateOf(false)
+    var context: Context? = null
+    val alice = User("uid-alice", "Alice", "alice@gmail.com", 42, emptyList(), "")
+    userViewModel.setCurrentUser(alice)
+    val firebaseAuth = mock(FirebaseAuth::class.java, RETURNS_DEFAULTS)
+    val currentUser = mock(FirebaseUser::class.java)
+    whenever(firebaseAuth.currentUser).thenReturn(currentUser)
+
+    composeTestRule.setContent {
```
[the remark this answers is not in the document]
SaturneV: I completely agree with you, I tried to perform the content setting in a startup function except that the setting screen composable depends on whether the user is null or not and displays two completely different screens depending on the case.

If I set the content in a startup function then it's not possible to recompose the function with a different user, so the test that tests the screen if the user is null fails every time given that the other tests need a non-null user, or vice-versa.

If you have a solution to this problem, don't hesitate to tell me about it, or if you have any questions, I'll be glad to help.

SaturneV:
Thanks for this detailed review!  I've added details on how to test the feature in the description according to your suggestion. I'll let you have a look at my answers to your comments, don't hesitate to ask for more clarifications.

app/src/androidTest/java/com/android/streetworkapp/ui/profile/SettingsTest.kt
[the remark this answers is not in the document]
misterM125: I see what you mean, i had not realized that this would be a problem, as I can't think of a better solution i think it is fine to leave it as is

misterM125 (approved):
Great changes. the delete account feature now works perfectly, thank you for applying the changes!

PR #224, opened 2024-12-12
Feat/implement UI for event filtering

Simmanz:
## Description

This PR aims to create a pop-up in which the user can set filters for the parks in the map.

#### Park filters
- By **Rating** => user specify the minimal rating in **[1, 5] stars**.
- By event density (thresholds for number of events) => {`LOW`, `MEDIUM`, `HIGH`}

#### New components

- `class ParkFilter`, which handle filtering using FilterSettings.
- `class FilterSettings`, which store the value of each park filter setting.
- **@Composable** `fun ParkFilterSettings`, which is the content of the pop-up, an interactible UI for settings of the filter.

### Big issues

I encountered MANY issues with the EventViewModel.

1. Initialy called eventViewModel.getEvent(park) then uiState => but it was **always** `uiState.Empty`

3. I implemented two new eventViewModel functions `getEventsByParkList()` which calls `getEventsByEid(eid_list)`
Now that the new implementation is called, the **eventList remains empty all the time** !

4. When going into `ParkOverview()`, **exiting it will cause parks with events to have their event list loaded**.
Essentially, the filters worked **only after going** into `ParkOverview()` once, becacuse the event list would remain empty otherwise.

5. I merged main with this branch, causing the `ParkOverview()` to no longer display the event list. Some calls are made which correctly load the event list, but the filters are only given empty lists. I am thinking of rolling back.

Not enough time remains to fix the eventViewModel bizarre behavior, nor the main branch breaking changes. Choose to remove the following filtering features : 

- By the status of the events => {`CREATED`, `ONGOING`, `FINISHED`}
- If spots remains for the user (#participants < #max)

### Tests

- Implemented unit test for the `FilterSettings` / `ParkFilter` API
- Implemented UI test for the filter settings user interface.

### Testing the feature :

1. New icon in the topAppBarActions in the `Screen.MAP`.
2. It should open a pop-up, containing rating and event density parameters. As well as a reset button.
3. Confirming should apply the filter settings.
4. Clicking outside the pop-up should discard changes.
5. Parks in the map should be filtered accordingly (they appear only if they match the filters)


app/src/main/java/com/android/streetworkapp/utils/ParkFilter.kt
```diff
@@ -0,0 +1,153 @@
...
+ * @property eventDensity The required event density.
+ */
+class FilterSettings {
+  val minRating: MutableState<Int> = mutableIntStateOf(DEFAULT_RATING)
+  val eventDensity: SnapshotStateList<EventDensity> = DEFAULT_EVENT_DENSITY.toMutableStateList()
+
+  /** The default filter settings. */
+  companion object {
+    val DEFAULT_RATING = 1
+    val DEFAULT_EVENT_DENSITY = listOf(EventDensity.LOW, EventDensity.MEDIUM, EventDensity.HIGH)
```
misterM125, line 63: `DEFAULT_RATING` should be a constant

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -169,6 +188,84 @@ fun MapScreen(
...
+          selected = userFilterInput.eventDensity.contains(EventDensity.MEDIUM),
+          onClick = { userFilterInput.updateDensity(EventDensity.MEDIUM) },
+          label = { Text(EventDensity.MEDIUM.name) },
+          colors = filterChipColors,
+          modifier = Modifier.padding(end = 2.dp).testTag("mediumDensityFilterChip"))
+
+      FilterChip(
+          selected = userFilterInput.eventDensity.contains(EventDensity.HIGH),
+          onClick = { userFilterInput.updateDensity(EventDensity.HIGH) },
+          label = { Text(EventDensity.HIGH.name) },
```
misterM125, line 252: For all three filterChips, putting LOW, MEDIUM, HIGH as the labels feels a bit vague for users to read, it might be nice to put the specific values as labels instead to help users better choose the density, for example: LOW (0 < ... < 2) or just (0 < ... < 2)

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
```diff
@@ -388,8 +390,11 @@ fun InteractiveRatingComponent(rating: MutableState<Int>) {
  * @param navigationActions The navigation actions to navigate to other screens.
  */
 @Composable
-fun EventItemList(eventViewModel: EventViewModel, navigationActions: NavigationActions) {
-  val uiState = eventViewModel.uiState.collectAsState().value
```
misterM125, line 392: I think this might be an error du to merges but the `uiState` in the parkOverview was declared directly inside the `EventItemList` because it is the only place where it is used, as such i would revert this change if it is  alright with you (i.e remove the uiState from the `EventItemList` and revert the above line)

app/src/test/java/com/android/streetworkapp/utils/ParkFilterTest.kt
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private fun createSimplerParks(rating: Float, events: List<String>): Park {
+    return Park(
+        "pid", "name", ParkLocation(0.0, 0.0, ""), "", rating, 0, 0, 0, events, emptyList(), "")
+  }
+
+  @Test
+  fun isParkFilterAPICorrect() {
+    val filterSettings = FilterSettings()
```
misterM125, line 61: `filterSettings` could be initialised as a global private value of the class instead of redeclaring it in every test

misterM125 (changes requested):
Great PR! 

The filters are a welcome addition to better help users narrow down specific parks.

I do have a couple suggestion to help make your code cleaner and make your filters easier and clearer to use for users

app/src/main/java/com/android/streetworkapp/utils/ParkFilter.kt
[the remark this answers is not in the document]
Simmanz: Actually the default filters are the following : 

- Minimum rating of 1 star (the minimum rating possible to give to a park)
- By default all kinds of event densities are accepted, giving an empty list would mean no park to display.

The sole thing which could be problematic is if when a park is added, its rating is initialized outside the allowed range [1, 5]. But I do not know how it is defined.

app/src/main/java/com/android/streetworkapp/utils/ParkFilter.kt
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Represents the filter settings.
+ *
+ * @property minRating The minimum rating required.
+ * @property eventDensity The required event density.
+ */
+class FilterSettings {
+  val minRating: MutableState<Int> = mutableIntStateOf(DEFAULT_RATING)
+  val eventDensity: SnapshotStateList<EventDensity> = DEFAULT_EVENT_DENSITY.toMutableStateList()
+
```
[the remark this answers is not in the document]
Simmanz: You can already access and modify filterSettings values. This can be seen in `ParkFilterSettings` or in the `Tests` :

First, you instantiate the settings : 
>val filterSettings = FilterSettings()

Then, use the setter to change the default values (It also supports giving only one filter argument)
>filterSettings.set(minRating = 4, eventDensity = listOf(EventDensity.LOW)

Finally, you can access them manually in such calls : 
>val minRating = filterSettings.minRating.value
val eventDensities = filterSettings.enventDensity.value

You can also modifiy them directly (albeit the setter is here to protect the inside values, for instance trying to set a minmum rating greater than 5 would not change the inside value) : 
>filterSettings.minRating.value = 2

app/src/main/java/com/android/streetworkapp/utils/ParkFilter.kt
Simmanz: This has been changed :)

app/src/main/java/com/android/streetworkapp/ui/park/ParkOverview.kt
Simmanz: Thanks for catching that ! It was done when trying to change / access the `eventViewModel` and should not have stayed !

app/src/test/java/com/android/streetworkapp/utils/ParkFilterTest.kt
Simmanz: I replaced the repetitive calls with a variable

app/src/main/java/com/android/streetworkapp/utils/ParkFilter.kt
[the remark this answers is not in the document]
misterM125: What i meant is that i think it would be better to make it that your density filters all start as unselected and accept all park densities in such a default state, then if a user clicks on HIGH density for example, then it will show only the high density parks, the other way around which you are currently doing feels a bit confusing to me, this might just be a detail, if you feel the filters are fine as is i'll still approve the review

app/src/main/java/com/android/streetworkapp/utils/ParkFilter.kt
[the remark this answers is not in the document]
misterM125: You are right, it's fine as it is now

app/src/main/java/com/android/streetworkapp/utils/ParkFilter.kt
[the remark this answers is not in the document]
Simmanz: I think that this is a bad idea, because if a user does not choose to select any density, then no park will be displayed.

The current implementation will display all of the parks if the user touches nothing. Then he can opt-out some parks.

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
Simmanz: I propose something in between as the ranges take too much space (reduced width). I came up with the following :

<p align="center">
image
</p>

Do you agree with such display ? It no longer give abstracted types but a direct idea of the range.

misterM125: This is great!

misterM125 (approved):
Great changes! 
Event filtering looks really good now

PR #225, opened 2024-12-12
fix: fixed erroneous string.xml use in park overview screen

misterM125:
This PR aims to resolve problems with the strings.xml used in the DateUtils of the park overview


Simmanz (approved):
Good changes ! You are now correctly using `strings.xml` calls for strings. Now that you have done it once, it will be easier for the next times (still one week left !).

PR #229, opened 2024-12-16
Feat/make icon change color in function of nb event

kzepfl:
## Feature
Icon color will go from : **blue (0 event) ==> pink (5+ event)**
<p align="center">
image
</p>

## Changes

- `MapUi`, made so icon change color depending on number of event and 2 helper function for the color gradient of the icon `gradientColor` and `colorToHue`
_Change `variable hotPlace` to change the number of event needed for maximum hotness_

## Testing

- `ColorTest` in Unite test, map folder. Test the helper function : `gradientColor` and `colorToHue`

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
+++ b/app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
@@ -108,7 +108,7 @@ fun MapScreen(
   }
 
   // Define how many event is considered a Hot place
-  val manyEvent = 3.0f
+  val hotPlace = 5.0f
 
   // Define the start and end colors for the gradient
   val startColor = ColorPalette.LOGO_BLUE
@@ -173,7 +173,7 @@ fun MapScreen(
 
                 // Interpolate color based on number of event (make the gradient)
                 interpolatedColor =
-                    gradientColor(startColor, endColor, (park.events.size) / manyEvent)
+                    gradientColor(startColor, endColor, (park.events.size) / hotPlace)
 
                 markerIcon = BitmapDescriptorFactory.defaultMarker(colorToHue(interpolatedColor))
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -106,6 +107,27 @@ fun MapScreen(
...
+  val endColor = ColorPalette.LOGO_RED
+
+  // variable for each park color
+  var interpolatedColor: Color // set default color
+  var markerIcon: BitmapDescriptor
+
+  // Handling user MVVM
+  val currentUser = userViewModel.currentUser.collectAsState().value
+
+  val userParkList = userViewModel.parks.collectAsState().value
```
alvaro1080, line 124: Suggestion: This userParkList does not seem to be used, it could be removed

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -222,3 +252,36 @@ fun MarkerInfoWindowContent(park: Park) {
         }
       }
 }
+
+fun gradientColor(startColor: Color, endColor: Color, fraction: Float): Color {
```
alvaro1080, line 256: Docs: the documentation is missing for this function

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -222,3 +252,36 @@ fun MarkerInfoWindowContent(park: Park) {
...
+
+  val r = startR + fraction * (endR - startR)
+  val g = startG + fraction * (endG - startG)
+  val b = startB + fraction * (endB - startB)
+  val a = startA + fraction * (endA - startA)
+
+  return Color(r, g, b, a)
+}
+
+fun colorToHue(color: Color): Float {
```
alvaro1080, line 275: Docs: same as the last comment, the documentation is also missing

alvaro1080 (changes requested):
Thank you for this Pull Request! In terms of UI, your Pull Request is particularly beautiful and the fact that the color are an echo to the logo is really appreciable. It reinforce the overall coherence of the app. On top of that, we have now an easy way to see the park with a lot of event which is good in terms of UX. 

I have a few details that I would like to highlight, but otherwise really good job!

alvaro1080 (approved):
Thank you for the update! I approve this Pull Request.

PR #239, opened 2024-12-18
Refactor/map UI

misterM125:
## Overview

This PR is a small refactor of map ui elements:
- The search bar is now always visible on map screen 
- A small "enter" icon has been added to marker info windows to enhance user experience

### Additional info
- some top app bar and map ui tests have been refactored to account for these changes

### DIFF
--- a/app/src/main/java/com/android/streetworkapp/MainActivity.kt
+++ b/app/src/main/java/com/android/streetworkapp/MainActivity.kt
@@ -255,6 +255,7 @@ fun StreetWorkApp(
   // Instantiate info manager and its components :
   val showInfoDialog = remember { mutableStateOf(false) }
   val showSearchBar = remember { mutableStateOf(false) }
+  val showFilterSettings = remember { mutableStateOf(false) }
 
   // query for the search bar
   val searchQuery = remember { mutableStateOf("") }
@@ -330,8 +331,6 @@ fun StreetWorkApp(
               route = Route.MAP,
           ) {
             composable(Screen.MAP) {
-              val showFilterSettings = mutableStateOf(false)
-
               infoManager.Display(LocalContext.current)
               MapScreen(
                   parkLocationViewModel,
--- a/app/src/main/java/com/android/streetworkapp/MainActivity.kt
+++ b/app/src/main/java/com/android/streetworkapp/MainActivity.kt
@@ -260,7 +260,7 @@ fun StreetWorkApp(
   // query for the search bar
   val searchQuery = remember { mutableStateOf("") }
 
-  Log.d("InfoDialog", "Main - Instantiating the InfoDialogManager")
+  Log.d("MainActivity", "Main - Instantiating the InfoDialogManager")
   val infoManager =
       InfoDialogManager(
           showInfoDialog, currentScreenName, topAppBarManager = screenParams?.topAppBarManager)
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/android/streetworkapp/MainActivity.kt
```diff
@@ -269,22 +268,15 @@ fun StreetWorkApp(
...
-                infoManager.setUp()
-              }
-        }
+        screenParams
+            ?.isTopBarVisible
+            ?.takeIf { it }
+            ?.let {
+              TopAppBarWrapper(navigationActions, screenParams?.topAppBarManager, searchQuery)
+              // setup the InfoDialogManager in topBar, because it relies on the topAppBarManager.
+              Log.d("InfoDialog", "Main - Setting up the InfoDialogManager")
```
tercierp, line 277: Suggestion: I know that it was already there, but for consistency, you could use the tag `MainActivity` instead of `InfoDialog`. The tag could be set as a constant at the beginning of the file.

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -290,54 +292,71 @@ fun ParkFilterSettings(userFilterInput: FilterSettings) {
...
-        colors = SearchBarDefaults.colors(containerColor = ColorPalette.INTERACTION_COLOR_LIGHT)) {}
-  }
+fun MapSearchBar(query: MutableState<String>) {
+  Row(
+      modifier = Modifier.fillMaxWidth().padding(vertical = 20.dp),
+      horizontalArrangement = Arrangement.Center) {
+        SearchBar(
+            query = query.value,
+            onQueryChange = { query.value = it },
+            placeholder = { Text("Search for parks") },
```
tercierp, line 302: Improvement: Since it's asked by the coaches, could you put this text in the strings.xml

I would do it like that:
```kotlin
Text(stringResource(R.string.park_search))
```
```xml
<string name="park_search">Search for parks</string>
```

app/src/main/java/com/android/streetworkapp/ui/map/MapUi.kt
```diff
@@ -290,54 +292,71 @@ fun ParkFilterSettings(userFilterInput: FilterSettings) {
...
+
+          RatingComponent(park.rating.toInt(), park.nbrRating)
+
+          Row(modifier = Modifier.padding(5.dp), verticalAlignment = Alignment.CenterVertically) {
+            Icon(
+                Icons.Filled.Event,
+                contentDescription = "eventIcon",
+                modifier = Modifier.padding(horizontal = 5.dp).testTag("eventIcon"))
+            Text(
+                text = "${park.events.size} event(s) planned",
```
tercierp, line 339: Improvement: This could also be in the `strings.xml`

tercierp (changes requested):
On my phone it looks really good, very good refactor ! I did find small details that could be improved.

misterM125:
Thank you for the suggestions, 

changes to the logs and `strings.xml` file have been applied

> ⚠️I didn't approve you because the `End2EndParks` no longer passes when you remove the `@Ignore` I think there must have been a number of changes that make this test no longer valid, can you make it pass at least locally?

I don't think i will apply these changes because 1. I might be wrong about that but this test has been dysfunctional for a while 
2. It was already deemed to be insufficient and invalid by the coaches and 3. every functionality that this this e2e intended to test is already tested in End2EndCreateEvent, as such refactoring this test would probably be a waste of time

tercierp (approved):

tercierp:
Ok I understand your point of view, thank you for your changes, it's approved.

PR #242, opened 2024-12-19
Feat/create end2end based on achievements

alvaro1080:
## Overview

The objective of this pull request is to create an end to end test based on achievements. In particular, we would like that this end2end works reliably on the CI, as this is important for the M3.


alvaro1080:
I did not manage to make a better end2end test than the one we already have, since this new test seem to fail on the CI like the ones we created before. I will close this branch since it does not improve the current state of the end2end test.

PR #244, opened 2024-12-19
fix(LocationService): fixed currentpark not updating by taking park d…

kzepfl:
## Fix

User current park didn't update, enabling people to spam point near a park by reloading the map
Fix: current user now is correctly updated, user won't get point when inside the same park

## Changes

- `LocationService.kt`, changed val currentParks to a var and updated currentPark by using the userViewModel directly

### You can test the functionnality by seting your location inside a park and reloading the map
image


Simmanz (approved):
Good fix ! I tested it : Refreshing the map is indeed no longer providing the points for discovering the park multiple times. Good job !

<p align="center">
image
</p>

PR #245, opened 2024-12-19
Feat/update the helping dialogs

Simmanz:
### Updated content : 

- `EDIT_EVENT` support - New `Strings.xml` value + added to the callback map : (`infoDialogManager` was already correctly used in main).
- `TRAIN_HUB` support - New `Strings.xml` value + added to the callback map + Manager call in `Mainactivity`
- Updated some values in `Strings.xml` to remove spaces after line breaks.

### How to test the feature

- `EDIT_EVENT` Screen should have a supported helping dialog.
- `TRAIN_HUB` Screen should have a supported helping dialog.
- Other helping dialogs should not have problematic formatting


kzepfl (commented):
Good PR, I checked the helping dialogue on my side and they all display

kzepfl (approved):
<p align="center">¨


image
image
image
image
</p>

